PR #24, opened 2015-02-13
Allow custom public suffix file

cicku:
There is an option already in the configure.ac to allow system-wide public suffix, Fedora ships the dat as a package "publicsuffix-list" and install it to /usr/share/. Thus I'd like to use the system one since it's updated often.

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -1,3 +1,3 @@
 filesdir = $(datadir)/@PACKAGE@
-files_DATA = effective_tld_names.dat test_psl.txt
+files_DATA = $(PSL_FILE) test_psl.txt
 EXTRA_DIST = $(files_DATA)


darnir:
Looks good to me and would be a great thing to have!

rockdaboot:
Thanks, cicku !
I merged your pull request (via command line) to branch 'develop' and will make a new release soon.
BTW, is there a 'trigger' in the Fedora packaging system, like 'if package A becomes rebuild, also rebuild package B' ? I would be nice to have libpsl rebuild automatically when 'publicsuffix-list' changes (or becomes rebuilt).

cicku:
@rockdaboot, we don't have such feature, I'm gonna request one.

BTW, does this library requires rebuilds when each time the suffixlist being updated?

darnir:
Yes, because libpsl does not dynamically read from from a PSL file, but has
all the contents stored in the binary

Thanking You,
Darshit Shah

rockdaboot:
libpsl is made to have the PSL data available to an app without I/O. The PSL data is built in (you can switch it off via ./configure option as well). So yes, Wget explicitely uses the built-in data. Thus you have to rebuild libpsl each time the suffixlist data changes.

cicku:
Yes, I understand.

I/O will consume more resources IMO, I and the publicsuffix-list packager will rebuild it.

Thanks.

PR #37, opened 2015-12-03
Fix Shebang line and other issues with autogen.sh

darnir:
The shebang line was not being interpreted due to a space between # and !. On
fixing it, smaller issues came to the fore due to the effects of the -e switch.
Fixed them all.

P.S.: If ou don't want GitHub pull requests, I can send you a patch over.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,21 +1,21 @@
-# !/bin/sh -e
+#!/bin/sh -e
 
-AUTORECONF=`which autoreconf 2>/dev/null`
+AUTORECONF=$(which autoreconf 2>/dev/null || true)
 if test $? -ne 0; then
   echo "No 'autoreconf' found. You must install the autoconf package."
   exit 1
 fi
 
-GIT=`which git 2>/dev/null`
+GIT=$(which git 2>/dev/null || true)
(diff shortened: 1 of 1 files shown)


darnir:
Both `` and $() are POSIX requirements. However, `` is kind-of deprecated and its use discouraged everywhere. $() is supported by sh, so it should work everywhere, except for 20+ year old implementations.

The main issue with the script wasn't ``, but the very first line. Adding a space between # and !, makes it a sh comment. While what you really wanted was #!/bin/sh which tells the OS to use /bin/sh to run the script.

I'm not sure about the portability of `mkdir -p` in this particular aspect. I think you're right, we should revert back to using a shell redirection. I'll update this patch.

rockdaboot:
Made a quick test on OpenBSD 5.3, looks good. Merged now, thanks !

rockdaboot:
+GTKDOCIZE=$(which gtkdocize 2>/dev/null || true)

This fails if gtkdocize does not exist (just had it on Solaris 11) ! It should IMO be || false. Or better leave it away - I can't see any sense in it at all.
Could you give a good explanation of the || true changes ?

PR #39, opened 2015-12-12
re-fix psl_builtin_outdated()

dkg:


rockdaboot:
*arg* you are right (just slapping myself ;-)

PR #45, opened 2016-01-04
Make sure make_dafsa.py uses Python2

darnir:
make_dafsa.py is a definitely a Python 2 script. On systems where python is symlinked to python3, this script fails causing the build process to fail.


rockdaboot:
Good To know. Do you see any chance to get the script working with python 2 + 3 ?

rockdaboot:
I fixed the python3 incompatibilities in branch 'develop'. Thanks for pointing it out.

PR #56, opened 2016-09-17
Change src/psl-make-dafsa shebang so it'll run on OS X

SteelPangolin:
OS X doesn't follow [PEP 394](https://www.python.org/dev/peps/pep-0394/) and thus it doesn't have a `python2` binary. However, `python` is always Python 2 on OS X, and should be new enough to run `psl-make-dafsa` on the last five or so releases of OS X.

If this would break compatibility with other non-PEP-394 systems where `python` might be Python 3, another alternative might be to use `python2.7` specifically.

### DIFF
--- a/src/psl-make-dafsa
+++ b/src/psl-make-dafsa
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # Copyright 2014 The Chromium Authors. All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE.chromium file.


SteelPangolin:
BTW, I've opened [a PR with a Homebrew formula for `libpsl`](https://github.com/Homebrew/homebrew-core/pull/4281), and this change was suggested by Homebrew reviewers.

rockdaboot:
Thanks, good point.
PEP-394 states that `python` points to 'python2'.

If a problem arises with python3 (non PEP-394 compliant), we care about it than. Keeping in mind that PEP-404 states, there will never be any 2.8 release and knowing that 2.6 won't work at all, using 'python2.7' in the shebang is a reasonable alternative.

But of course, someone could work on changing psl-make-dafsa to work on both, 2.7. and 3.x.
I spent an hour or so, giving up after introducing more and more python version checks, making the code ugly. Not worth it since python2.7 will be there for a long time...

PR #69, opened 2016-11-12
Fix psl-make-dafsa to work with python3

liljenzin:
String encoding is quite different in python3 and I had to use bytes objects instead.

"make check" will now pass with both python2 and python3, otherwise I have not made much testing.


rockdaboot:
Wow ! @darnir said it's a nontrivial change... I didn't expect that vast. I'll do all kind of testings tomorrow. Thanks again, Olle !

rockdaboot:
Many thanks, passes all tests. python2 (2.7.12) python3 (3.5.2) create exactly the same binaries here.

PR #79, opened 2017-08-29
Add new function psl_free_string()

darnir:
When writing a wrapper around LibPSL in a different language it is
important that libpsl provide functions to free any memory that it
allocates. Without this, it is impossible to correctly free the memory
allocated by psl_str_to_utf8lower() function since in other languages
one may not have access to the same free() call from libc.

### DIFF
--- a/src/psl.c
+++ b/src/psl.c
@@ -1614,6 +1614,21 @@ int psl_is_cookie_domain_acceptable(const psl_ctx_t *psl, const char *hostname, 
 }
 
 /**
+ * psl_free_string:
+ * @str: pointer to lowercase string returned by psl_str_to_utf8lower()
+ *
+ * This function free()'s the memory allocated by psl_str_to_utf8lower() when
+ * returning a lowercase string
+ *
+ * Since: 0.19
+ */
+void psl_free_string(char *str)
+{
+	if (str)
+		free(str);
+}
+
+/**
  * psl_str_to_utf8lower:
  * @str: string to convert
  * @encoding: charset encoding of @str, e.g. 'iso-8859-1' or %NULL
--- a/include/libpsl.h.in
+++ b/include/libpsl.h.in
@@ -79,6 +79,10 @@ typedef struct _psl_ctx_st psl_ctx_t;
 void
 	psl_free(psl_ctx_t *psl);
 
+/* frees memory allocated by libpsl routines */
+void
+	psl_free_string(char *str);
+
 /* loads PSL data from file */
 psl_ctx_t *
 	psl_load_file(const char *fname);
--- a/tests/test-is-public.c
+++ b/tests/test-is-public.c
@@ -133,16 +133,16 @@ static void test_psl(void)
 		char *lower = NULL;
 
 		psl_str_to_utf8lower("www.example.com", NULL, "de", &lower);
-		free(lower); lower = NULL;
+		psl_free_string(lower); lower = NULL;
 
 		psl_str_to_utf8lower("\374bel.de", NULL, "de", &lower);
-		free(lower); lower = NULL;
+		psl_free_string(lower); lower = NULL;
 
 		psl_str_to_utf8lower("\374bel.de", "iso-8859-1", NULL, &lower);
-		free(lower); lower = NULL;
+		psl_free_string(lower); lower = NULL;
 
 		psl_str_to_utf8lower(NULL, "utf-8", "en", &lower);
-		free(lower); lower = NULL;
+		psl_free_string(lower); lower = NULL;
 	}
 
 	psl_get_version();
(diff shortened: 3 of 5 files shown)


src/psl.c
```diff
@@ -1614,6 +1614,21 @@ int psl_is_cookie_domain_acceptable(const psl_ctx_t *psl, const char *hostname,
...
+ * @str: pointer to lowercase string returned by psl_str_to_utf8lower()
+ *
+ * This function free()'s the memory allocated by psl_str_to_utf8lower() when
+ * returning a lowercase string
+ *
+ * Since: 0.19
+ */
+void psl_free_string(char *str)
+{
+	if(str)
```
daurnimator: The check shouldn't be needed (`free` already does it internally)

darnir:
* daurnimator <notifications@github.com> [170830 02:50]:
>daurnimator commented on this pull request.
>
>
>
>> @@ -1614,6 +1614,21 @@ int psl_is_cookie_domain_acceptable(const psl_ctx_t *psl, const char *hostname,
> }
>
> /**
>+ * psl_free_string:
>+ * @str: pointer to lowercase string returned by psl_str_to_utf8lower()
>+ *
>+ * This function free()'s the memory allocated by psl_str_to_utf8lower() when
>+ * returning a lowercase string
>+ *
>+ * Since: 0.19
>+ */
>+void psl_free_string(char *str)
>+{
>+	if(str)
>
>The check shouldn't be needed (`free` already does it internally)

`free` in glic does the check internally, but I'm not sure if other libc 
implementations also provide the check. I think there are a few 
implementations where the libc free() does not check for a null pointer. 

Anyays, on a modern system, when compiling with LTO, this check should 
be optimized out by the compiler, but provides a safety net for when 
`free` does not test

>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/rockdaboot/libpsl/pull/79#pullrequestreview-59421097

-- 
Thanking You,
Darshit Shah
PGP Fingerprint: 7845 120B 07CB D8D6 ECE5 FF2B 2A17 43ED A91A 35B6

daurnimator:
> `free` in glic does the check internally, but I'm not sure if other libc implementations also provide the check.

According to the C standard they MUST. From the C1x standard section 7.22.3.3 (similar clauses can be found in older standards):
> If ptr is a null pointer, no action occurs

> Anyays, on a modern system, when compiling with LTO, this check should be optimized out by the compiler,

*very* few systems currently turn on LTO. They're a rarity.
For the majority of systems you've just added an extra branch.

src/psl.c
rockdaboot: It doesn't hurt and we do those checks in the library as well (but maybe not consistent everywhere).

@darnir, could you add a space between `if` and `(`, please !?

rockdaboot:
Thanks Darshit. Without it, we definitely have a problem on Windows. Remember those reports from Gisle Vanem ? :-)

rockdaboot:
@daurnimator This is a C89 library and I am not sure how old some build systems might. I personally remember the times where free(NULL) just crashed. Well, we could look into gnulib and follow their rules - the maintainer, Bruno Haible, is an absolutely expert and knows more about different systems than anyone else I ever heard of.

daurnimator:
> @daurnimator This is a C89 library and I am not sure how old some build systems might.

Even in C89 `free(NULL)` is fine. http://port70.net/~nsz/c/c89/c89-draft.html#4.10.3.2

> I personally remember the times where free(NULL) just crashed

Last one I know of was the early versions of the palm pilot toolchain.
Even console tool-chains from the 90s allowed it.

rockdaboot:
> Even in C89 free(NULL) is fine.

What if someone comes up with "free(NULL) crashes on my ancient <whatever> system" ? Should we say "LMA, we give a sh*t on this" or should we be polite and say "Ok, no prob - we add a check" ?

If the second, then let's do it right now !

daurnimator:
> What if someone comes up with "free(NULL) crashes on my ancient system" ? Should we say "LMA, we give a sh*t on this" or should we be polite and say "Ok, no prob - we add a check" ?

Then it's probably time to upgrade that ancient system. Supporting *pre-C89* is a fools errand.

rockdaboot:
> Then it's probably time to upgrade that ancient system. Supporting pre-C89 is a fools errand.

No, it's not when you are doing this for fun, research, just out of interest or whatever. Especially here in Germany people like to collect old stuff (hardware & software) to play around with it. It doesn't get the point telling them to upgrade. I also have customers with hardware older than 20 years (stil running some SCO Unix). They can't upgrade because they use special hardware, not available or not affordable today, simply not replaceable - and no Linux drivers at all. The SCO compiler there is pre-C89... but they need some software every several years compiled/build with these tools.

darnir:
> Thanks Darshit. Without it, we definitely have a problem on Windows. Remember those reports from Gisle Vanem ? :-)

I had a problem when trying to wrap the library in Rust. Since rust uses `jemalloc` instead of malloc().

Also, I think a single conditional shouldn't really be a performance impact in any way. On x86 this should compile into a single machine instruction and one that the CPU's branch predictor should be capable of handling.

PR #80, opened 2017-09-12
move psl-make-dafsa to python3

dkg:
Modern systems shouldn't need python 2.  psl-make-dafsa should work
with py3 just fine, as long as we know the source file is UTF-8.

I've shipped this in debian unstable now, since psl-make-dafsa was one of the few things on my system that was preventing me from purging python 2 on systems i care about.

It'd be great to adopt this change upstream.

### DIFF
--- a/src/psl-make-dafsa
+++ b/src/psl-make-dafsa
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # Copyright 2014 The Chromium Authors. All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE.chromium file.
@@ -685,7 +685,7 @@ def main():
     psl_nexceptions = 0
     psl_nwildcards = 0
 
-    with open(sys.argv[-2], 'r') as infile, open(sys.argv[-1], 'wb') as outfile:
+    with open(sys.argv[-2], 'r', encoding="utf-8") as infile, open(sys.argv[-1], 'wb') as outfile:
       outfile.write(converter(parser(infile, utf_mode, codecs), utf_mode, codecs))
 
   return 0


liljenzin:
I thought the command python was linked to the system default python, so that the shebang without version number would run python3 on the modern systems you refer to. Is this wrong? I don't have any python3-only machine to test on.

rockdaboot:
> I thought the command python was linked to the system default python, so that the shebang without version number would run python3 on the modern systems you refer to.

This is IMO right. The shebang should have no version number.

The encoding seems ok, assumed that we document that input files must be utf-8. @liljenzin You are the python expert - WDYT ?

rockdaboot:
We could also do this trick (see answers) and prefer python3 over python: https://stackoverflow.com/questions/18993438/shebang-env-preferred-python-version

rockdaboot:
Also see https://www.python.org/dev/peps/pep-0394/, `python` links to python3 on Arch and in the future it seems likely that more distros follow this behavior. Scripts written for python2 and python3 (as psl-make-dafsa) should use `python` without version in their shebang line.

liljenzin:
As long as the script generates identical output with both engines, I see no point in preferring one version before the other. Using system default python, as we currently do, makes the script runnable on most systems without adding dependency on a python engine that might not be installed by default.

Opening the input file with encoding utf-8 might be the right thing to do in python3, but the encoding argument is not supported in python2. If used it needs be conditional on the actual python version to not break python2 compatibility.

dkg:
I know of no plans on Debian (the operating system i use) to make
/usr/bin/python point to python3.

Since i want psl-make-dafsa to ship as an independent package on debian
(to facilitate cross-compilation) and i don't want psl-make-dafsa to
cause a dependency on python 2.x, i'm shipping it with this patch in
the debian package anyway.

some of the debian build daemons are not configured with a default
locale that is UTF-8 -- that's why the encoding argument is necessary in
python3.

It sounds like psl-make-dafsa is probably already broken on Arch systems
that do not have a UTF-8 locale.

If you prefer to just apply a version-flexible update to handle the
encoding parameter, then the debian-specific patch will consist of only
changing the shebang line, which is fine with me.

Thanks for looking at this!

       --dkg

rockdaboot:
@liljenzin I'm not deep into python, do you see an elegant way for compatibilty ?

Already tried (without success):
```
from io import open
...
with open(sys.argv[-2], 'r', encoding='utf-8') ...
```

and
```
    open_args = [ sys.argv[-2], 'r' ]
    if sys.version_info.major > 2:
      open_args.extend([ 'encoding=\'utf-8\'' ])

    with open(*open_args) as infile, open(sys.argv[-1], 'wb') as outfile:
```

liljenzin:
```
-    with open(sys.argv[-2], 'r') as infile, open(sys.argv[-1], 'wb') as outfile:
+    with open(sys.argv[-2], 'r', **codecs) as infile, open(sys.argv[-1], 'wb') as outfile:
```

rockdaboot:
@dkg Just tested and pushed @liljenzin's suggestion. Could you test it and report back, please. If it works for you (together with your python3 shebang), I could make a release during the next days.

dkg:
yep, i think it works for me.  thanks for this!

rockdaboot:
Thanks to both of you for your input !

PR #81, opened 2017-10-27
remove leading underscores, and remove fread/realloc loop in psl_load_fp()

alevesely:
I thin I removed all the leading underscores in C source.  That's because the C standard (at least some versions of it) reserve them for compiler implementation.  This is not what I meant to change, but I couldn't stand them, so I removed.  This pull request consists of two commit, the first one is about the underscores.

The second commit is removes a stepwise read loop.  It's easier to slurp the whole file at once.  This diff should be readable.  It would be even better to use <code>mmap()</code>, I'll try next week...

I just run <code>make</code> and <code>make check</code>.  I hope that exercised <code>psl_load_fp()</code>.

### DIFF
--- a/include/libpsl.h.in
+++ b/include/libpsl.h.in
@@ -73,7 +73,8 @@ typedef enum {
 	PSL_ERR_NO_MEM = -6    /* failed to allocate memory */
 } psl_error_t;
 
-typedef struct _psl_ctx_st psl_ctx_t;
+struct psl_ctx_st;
+typedef struct psl_ctx_st psl_ctx_t;
 
 /* frees PSL context */
 void
--- a/src/lookup_string_in_fixed_set.c
+++ b/src/lookup_string_in_fixed_set.c
@@ -8,15 +8,15 @@
 #include <stddef.h>
 
 #if defined(__GNUC__) && defined(__GNUC_MINOR__)
-#       define _GCC_VERSION_AT_LEAST(major, minor) ((__GNUC__ > (major)) || (__GNUC__ == (major) && __GNUC_MINOR__ >= (minor)))
+#       define GCC_VERSION_AT_LEAST(major, minor) ((__GNUC__ > (major)) || (__GNUC__ == (major) && __GNUC_MINOR__ >= (minor)))
 #else
-#       define _GCC_VERSION_AT_LEAST(major, minor) 0
+#       define GCC_VERSION_AT_LEAST(major, minor) 0
 #endif
 
-#if _GCC_VERSION_AT_LEAST(4,0)
-#  define _HIDDEN __attribute__ ((visibility ("hidden")))
+#if GCC_VERSION_AT_LEAST(4,0)
+#  define HIDDEN __attribute__ ((visibility ("hidden")))
 #else
-#  define _HIDDEN
+#  define HIDDEN
 #endif
 
 #define CHECK_LT(a, b) if ((a) >= b) return 0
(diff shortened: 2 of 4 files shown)


rockdaboot:
Hi Alessandro,

thanks for your contribution.

The C standard proposes to use _ for file scope identifiers.
Here's what the C standard says (section 7.1.3):

- All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use.
- All identifiers that begin with an underscore are always reserved for use as identifiers with file scope in both the ordinary and tag name spaces.

The read loop works with chunks of 64k by purpose. There is no sense in allocating memory as huge as a file is. And 65k is large enough to read the standard PSL in one chunk. This might change in the future as more and more entries are added. But it's fine for now.

And there is no need for mmap here. First, it is not portable and thus you need a fallback anyways. Second, there is no speed bonus for small files like we have. If you load gigabytes of files, there might be a performance win. But my benchmarks never showed anything remarkable on current Linux. So please don't do it.

There are certain things that I would like to take from your code (or better: you come up with a new commit). That is the removal of _ before defines (e.g. _HIDDEN) - that in fact isn't C standard.

Before you work on any code (true for almost any project), open an issue with a description and that you are willing to care about it. So the community and/or maintainer(s) can discuss it and you won't waste your time. The issue becomes assigned to you so everybody knows you are working on it.

alevesely:
On Fri 27/Oct/2017 23:07:07 +0200 Tim Rühsen wrote:
> 
> The C standard proposes to use _ for file scope identifiers.
> Here's what the C standard says (section 7.1.3):
> 
>   * All identifiers that begin with an underscore and either an uppercase
>     letter or another underscore are always reserved for any use.
>   * All identifiers that begin with an underscore are always reserved for use
>     as identifiers with file scope in both the ordinary and tag name spaces.

Don't those words mean that those identifier are reserved?  Most of them are
static functions, for which there's no reason to use a reserved name.

I understand that such changes are messy.  Yet, If you know a clean way to
carry them out, the sooner the better.

> The read loop works with chunks of 64k by purpose. There is no sense in
> allocating memory as huge as a file is. And 65k is large enough to read the
> standard PSL in one chunk. This might change in the future as more and more
> entries are added. But it's fine for now.

I allocated 16 bytes more than needed, is that what you mean?

What is the reason to allocate a whole segment and then adjust its size downward?

And, since I'm at it, what is the reason of not keeping that 16 byte header in
memory, as well as in kDafsa?  There is room for a 32-bit date there (keeping
just one byte for format '0'), and a date saved inside the file could be more
reliable than the file date, as the latter may change on copying or on downloading.

Just ruminating...

> And there is no need for mmap here. First, it is not portable and thus you need
> a fallback anyways. Second, there is no speed bonus for small files like we
> have. If you load gigabytes of files, there might be a performance win. But my
> benchmarks never showed anything remarkable on current Linux. So please don't
> do it.

The idea was to have a single file in memory, shared among all users of the
library.  That's the way memory is used by kDafsa.  However, the compiled-in
data is bound to become obsolete quickly.  Therefore, methink, it would be more
reliable to routinely load a supposedly updated file.

> Before you work on any code (true for almost any project), open an issue with a
> description and that you are willing to care about it. So the community and/or
> maintainer(s) can discuss it and you won't waste your time. The issue becomes
> assigned to you so everybody knows you are working on it.

Yes, I shouldn't have tried to fix libpsl, since it is not broken :-)  In fact,
I did file a bug, but not about libpsl in particular:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=879008

I didn't know how libpsl manages updates, so I began to browse the sources to
learn it.  Then, I just couldn't resist...


Thank you for your prompt reply, anyway
Ale
--

rockdaboot:
Sorry, not enough time to discuss everything in detail (not that I wouldn't like to).

> Yes, I shouldn't have tried to fix libpsl, since it is not broken

Your changes are mostly "personal preference" changes (except the mentioned _[A-Z] identifiers). If your commits would have been cleaner (one commit per task), I could simply apply one of them. Now I have to put time into extracting the relevant pieces - time that I hardly have. If you are going to make up a new PR with only those changes, that would be fine.

PR #89, opened 2018-01-16
GCC flags

dago:
Flags specific to GCC should not be set unconditionally


rockdaboot:
Thanks, pushed from local without that "Test2".

PR #90, opened 2018-02-21
psl_is_public_suffix2(): allow checking for suffixes not in the list

csaavedra:
Add a PSL_TYPE_NO_STAR_RULE type to check for suffixes without the '*'
rule. This allows checking for suffixes that are not in the PSL.


rockdaboot:
Great ! LGTM :-)

Travis mourns, because i added a C99 feature recently and this library is pure C89 (to build on legacy platforms).  I just pushed a fix.

Adding tests would be needed to see if it's working as expected.
I suggest to extend test-is-public.c and test-is-public-builtin.c in the following way:
- Add a second result (e.g. result2) to `struct test_data`.
- Copy the loop and call `psl_is_public_suffix2` with the new flag or'ed with the ANY flag and amend the if construct to test result2.
- Optional: Add some more tests to the array as you like.

And then rebase to see Travis CI succeed.

csaavedra:
The tests will fail for a handful of cases; but I think the expectations I wrote are correct. This might be related to the fact that there is a new codepath that might not have been tested before. But please, correct me if the expectations I added and are failing are not correct. Same if there are any other added expectations that are wrong.

rockdaboot:
Fix some smaller things in your last commit and added TLDs to the data. Before we skipped them because of the star rule, so the new tests failed on single TLDs from the PSL.

I will make up a new release in the next days.

Thanks again for your contribution !

PR #97, opened 2018-04-18
Make code buildable on native Windows (Win32)

fanc999:
Hi,

As GNOME's libsoup, which is supported on Windows natively, now depends on the libpsl library, I am proposing some simple changes to the code so that libpsl can build and run on Windows without the need for Cygwin, and so that libsoup will continue to work natively on Windows.

I will also (possibly in another PR) contribute a set of NMake Makefiles to build libpsl under Visual Studio.

With blessings, thank you!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -122,25 +122,48 @@ AC_ARG_ENABLE([asan],
 AC_SUBST([LIBPSL_SO_VERSION], [8:0:3])
 AC_SUBST([LIBPSL_VERSION], $VERSION)
 
+#
+# Check for Win32
+#
+
+AC_MSG_CHECKING([for Win32])
+case "$host" in
+  *-*-mingw*)
+    os_win32=yes
+    ;;
+  *)
+    os_win32=no
+    ;;
+esac
+AC_MSG_RESULT([$os_win32])
+AM_CONDITIONAL(OS_WIN32, [test $os_win32 = yes])
+
+# Windows does not have langinfo.h, which the libidn/libidn2
+# builtin and runtime need, so default to ICU
+if test "x$os_win32" = "xyes" ; then
+  default_runtime=libicu
+  default_builtin=libicu
+else
+  default_runtime=libidn2
+  default_builtin=libidn2
+fi
+
 # Check for enable/disable builtin PSL data
 AC_ARG_ENABLE(runtime,
   [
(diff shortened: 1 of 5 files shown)


rockdaboot:
Hi, thanks for your contribution !

A question regarding the decorator, shouldn't it look more like the following ? Can you test that on Windows ?
```
// see https://www.gnu.org/software/gnulib/manual/html_node/Exported-Symbols-of-Shared-Libraries.html
#if HAVE_VISIBILITY
#       define PSL_API __attribute__ ((__visibility__("default")))
#elif defined _MSC_VER && !defined PSL_STATIC
#       define PSL_API __declspec(dllexport)
#elif defined _MSC_VER && !defined PSL_STATIC
#       define PSL_API __declspec(dllimport)
#else
#       define PSL_API
#endif
```

fanc999:
Hi @rockdaboot,

It could, but this was done this way so that we could define that as a CFLAG or a config.h variable during the build of the library.  This essentially follows the pattern of what is now done in libsoup, GLib and so, in GNOME, albeit that I didn't define that as "extern" by default.  We could probably then use some autotools (or so) definition, like what is done in GLib to define PSL_API accordingly in config.h, which will do what is needed during the build of libpsl.

The above would probably not be enough for _MSC_VER because there needs to be a distinction when we are building libpsl and when we are using libpsl (for instance, we need to have a CFLAG that says something like -DPSL_COMPILATION or so when we are building libpsl to use __declspec(dllexport)).

Let me know which way is preferable though, so that I can make the necessary changes, if needed.

With blessings, thank you!

rockdaboot:
Is this more appropriate ? I'd prefer something like this, having other projects in mind. If you need (only if), we could also have a `#ifdef PSL_API` around it.
```
#if defined BUILDING_PSL && HAVE_VISIBILITY
#       define PSL_API __attribute__ ((__visibility__("default")))
#elif defined BUILDING_PSL && defined _MSC_VER && !defined PSL_STATIC
#       define PSL_API __declspec(dllexport)
#elif defined _MSC_VER && !defined PSL_STATIC
#       define PSL_API __declspec(dllimport)
#else
#       define PSL_API
#endif
```

rockdaboot:
@fanc999 I merged your code manually and added another commit to extend and rename the decorator to PSL_API. Please check if that works for you now.

PR #98, opened 2018-04-19
Enable autotools builds to build Windows binaries

fanc999:
Hi,

Apparently I was working on my branch at the same time when it was being merged, sorry!  The other changes that was done in libpsl.h.in shouldn't affect things at this point.

This updates the autotools builds (which I am not too good at) for enabling builds with MinGW.  I will also work on Visual Studio NMake Makefiles shortly afterwards.

With blessings, thank you!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -22,7 +22,7 @@ ACLOCAL_AMFLAGS = -I m4 ${ACLOCAL_FLAGS}
 pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = libpsl.pc
 
-EXTRA_DIST = config.rpath LICENSE
+EXTRA_DIST = build-aux/config.rpath LICENSE
 dist-hook:
 	mkdir -p $(distdir)/list/tests
 	cp -p $(PSL_FILE) $(distdir)/list


rockdaboot:
The commit is too windows-centric. So here are some questions:

> We link to -lws2_32, which is necessary for the networking functions on Windows.

Why don't you use `AC_SEARCH_LIBS(inet_ntop, ws2_32)` or something like this ?

> We default to using libicu on Windows...

That's what we have configure flags for. Please describe the problem you have using them.

> ...not have fmemopen()

So please add fmemopen to AC_CHECK_FUNCS in configure.ac, and modify the fuzzer code to exit(77) if HAVE_FMEMOPEN is not defined. That displays a 'SKIP' when running the test suite.

fanc999:
Hi Tim,

I updated the check for inet_pton(), which was what was being used, as you suggested.

For the libicu part, I am leaving that intact for now, as basically the issue is where we set the default runtime and builtin as libidn2.  The thing is code that uses libidn2 (and libidn as well) make use of items in langinfo.h, which is a header that is not available on native Windows builds (be it MinGW or Visual Studio).  This means that --enable-runtime=libicu and --enable-builtin=libicu is required for the code to build on Windows at the current state (or we disable the runtime annd builtins).  Another way is to update README.md to tell people that currently it is required to pass in --enable-runtime=libicu and --enable-builtin=libicu (or use --disable-runtime and --disable-builtin) to build libpsl on Windows.

Please let me know which is your preferred way to this issue.

Also, I am doing another commit here for the fuzz tests part.

With blessings, thank you!

rockdaboot:
> make use of items in langinfo.h

Right, there also should be a guard around the cal to `nl_langinfo()` and in the else case call the appropriate Windows function. Do you have it at hand ? I have it somewhere, must search...

> Another way is to update README.md to tell people that currently it is required to pass in --enable-runtime=libicu and --enable-builtin=libicu (or use --disable-runtime and --disable-builtin) to build libpsl on Windows.

That sounds good for now. On the other hand, if libidn2 development files aren't installed, ./configure should fallback to libidn and then to libicu automatically. I thought it was working that way, but will have a look. Maybe it works if you fix the nl_langinfo() thing ?

fanc999:
Hi Tim,

I will look into nl_langinfo() with the appropriate Windows replacement later.

For the default runtime/builtin, unfortunately configure errors out, saying something like requested libidn2 but could not find it, so no automatic fallback is being done there.

I think for the Visual Studio builds, getting libidn2/libidn and libunistring is going to be tedious, due to their use of gnulib (big autotools-using stuff there)...  So, it is likely that for the Visual Studio builds, I will focus on using ICU and the case where the runtime/builtin is disabled.

With blessings, thank you!

rockdaboot:
OK, just let me know if you are interested in a MinGW build script for libunistring, libidn2 and libidn. I use one for wget2 CI builds. Might help before you start from scratch.

rockdaboot:
@fanc999 Pushed two of your commits and added another one implementing the fallback thing I mentioned. Indeed, it wasn't implemented (must have been a dream ;-)). Thanks for your work.

fanc999:
Hi,

I posted a new patch to enable DLL builds on Windows, as we need the -no-undefined linker flag on Windows for building DLLs, and enabled linking to libintl and libiconv when we are using the libidn/libidn2 code, otherwise the build will fail to link since APIs from these two libraries are being used (libintl and libiconv are things that are not in the MinGW/Windows CRT library).

Let me know if these two changes could be done in a better way, since I am by no means good in autotools.

With blessings, thank you!

rockdaboot:
Can you remove the OS specific code and try this instead, please:
```
diff --git a/configure.ac b/configure.ac
index f0e627b..501d83e 100644
--- a/configure.ac
+++ b/configure.ac
@@ -10,7 +10,7 @@ AC_CONFIG_HEADERS([config.h])
 AC_PROG_CC
 m4_ifdef([AM_PROG_AR], [AM_PROG_AR])
 #LT_INIT([disable-static])
-LT_INIT
+LT_INIT([win32-dll])
 AC_CONFIG_MACRO_DIR([m4])
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
diff --git a/src/Makefile.am b/src/Makefile.am
index e18d754..c645b70 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -11,7 +11,7 @@ libpsl_la_CPPFLAGS = -I$(top_srcdir)/include -I$(top_builddir)/include -DPSL_DIS
   $(CFLAG_VISIBILITY) -DBUILDING_PSL
 
 # include ABI version information
-libpsl_la_LDFLAGS = -version-info $(LIBPSL_SO_VERSION)
+libpsl_la_LDFLAGS = -no-undefined -version-info $(LIBPSL_SO_VERSION)
 if WITH_LIBICU
   libpsl_la_LDFLAGS += $(LIBICU_LIBS)
 endif
```

It's from https://www.gnu.org/software/libtool/manual/html_node/LT_005fINIT.html, scroll down to 'win32-ddl'. The code works here. If it works for you we should use it.

The second issue, libintl and libiconv, should also be done automatically by libtool. We just have to add `$(LTLIBICONV) $(LTLIBINTL) @INTL_MACOSX_LIBS@` in Makefile.am at the right place(s). I'll try to find a machine/OS for testing.

rockdaboot:
Added two commits that address the above. Please check and give me feedback.

fanc999:
Hi Tim,

The two commits do work.

With blessings, thank you!

PR #99, opened 2018-04-26
Add NMake Makefiles for Visual Studio builds

fanc999:
Hi,

This adds a set of NMake Makefiles and associated items for building libpsl using Visual Studio's NMake.  This also pulls out the source listings of libpsl into a separate Makefile snippet so that it can be included  by both the autotools Makefile.am and the NMake Makefiles for the build.

This attempts to make things modularized and this reuses items that have been used in the past for building items such as HarfBuzz (which I maintained the NMake Makefiles for it before it switched to CMake), due to the differences between GNUish Make and NMake.  Due to the complexity of building libidn/libidn2 with libunistring using Visual Studio, currently only builds using libicu as the runtime/builtin, or builds without runtime/builtin usage are supported with this set of NMake Makefiles.

With blessings, thank you!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,7 +8,7 @@ if ENABLE_MAN
   SUBDIRS += docs/libpsl
 endif
 endif
-SUBDIRS += fuzz tests
+SUBDIRS += fuzz tests msvc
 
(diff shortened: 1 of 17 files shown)


rockdaboot:
Great, thank you !

PR #101, opened 2018-04-27
Use LIBIDN2_LIBS instead of hardcoding

dago:


dago:
@rockdaboot I rolled this forward but still get the same error, how can we proceed here? Can you reproduce the error on your Linux system?

rockdaboot:
@dago That's why I use hard-coded -lidn2 -lunistring ;-) Travis has an old version of libidn2, maybe even without pkg-config file. Same problem may arise with static builds.

What is your problem on Solaris ? You have built libunistring and libidn2, right ? static and shared... 
Else I have build scripts for both, working on Solaris as well, AFAIR. What is your fastest machine to test on ? I could give it a try at the weekend.

dago:
There is no libidn.so or libidn2.so natively in Solaris, so I also package them for OpenCSW and they are located in `/opt/csw/lib` and `/opt/csw/lib/64` which I set via `LDFLAGS=/opt/csw/lib`. This is honoured when I use `$(LDFLAGS)` but ignored when hardcoded as `-lidn`. Technically it is probably best to also test if `-lidn'` requires ' -lunistring` to produce a working binary and then conditionally add `-lunistring` to `$(LIBIDN)`.

The fastest machine is probably `unstable10x` or `unstable11s`.

rockdaboot:
Since you have libidn2 installed, on unstable11x the following works:
```
./autogen.sh
./configure
make
make check
```

Maybe I didn't understand your problem - building libpsl simply works on OpenCSW / unstable11x.

dago:
If It works for you because the default copiler that is picked up is `gcc` which comes also from `/opt/csw` directory hierarchy and implicitly links against stuff in `/opt/csw/lib`:

```
dam@unstable11x [global]:/home/dam/work/libpsl/src > gcc -shared  -fPIC -DPIC -Wl,-z -Wl,text -Wl,-h -Wl,libpsl.so.5 -o .libs/libpsl.so.5.3.1  .libs/libpsl_la-psl.o .libs/libpsl_la-lookup_string_in_fixed_set.o   -lidn2 -lunistring -liconv -lintl -lsocket -lnsl  -g -O2
dam@unstable11x [global]:/home/dam/work/libpsl/src > truss -f gcc -shared  -fPIC -DPIC -Wl,-z -Wl,text -Wl,-h -Wl,libpsl.so.5 -o .libs/libpsl.so.5.3.1  .libs/libpsl_la-psl.o .libs/libpsl_la-lookup_string_in_fixed_set.o   -lidn2 -lunistring -liconv -lintl -lsocket -lnsl  -g -O2 2>&1 | grep libidn2.so
28600:  open("/opt/csw/lib/gcc/i386-pc-solaris2.10/5.5.0/libidn2.so", O_RDONLY) Err#2 ENOENT
28600:  open("/opt/csw/lib/gcc/i386-pc-solaris2.10/5.5.0/../../../../i386-pc-solaris2.10/lib/libidn2.so", O_RDONLY) Err#2 ENOENT
28600:  open("/opt/csw/lib/gcc/i386-pc-solaris2.10/5.5.0/../../../libidn2.so", O_RDONLY) = 4
```

Try setting the compiler to something that does not explictly link to `/opt/csw/lib` e.g. `export CC=/opt/solarisstudio12.4/bin/cc`. The problem is that after setting `export LDFLAGS=-L/opt/csw/lib` then `libidn2.so` is properly detected during configure but the library path is not used during linking when the library is compiled with the Makefile rule omitting `$(LDFLAGS)`.

So to reproduce on `unstable11x`:
```
export CC=/opt/solarisstudio12.4/bin/cc
export LDFLAGS=-L/opt/csw/lib
./autogen.sh
./configure
make
make check
```
and it will fail with
```
gmake[3]: Entering directory '/home/dam/work/libpsl/src'
/bin/sh ../libtool  --tag=CC   --mode=link /opt/solarisstudio12.4/bin/cc  -g -no-undefined -version-info 8:1:3  -lidn2 -lunistring  -L/opt/csw/lib -o libpsl.la -rpath /usr/local/lib libpsl_la-psl.lo libpsl_la-lookup_string_in_fixed_set.lo -L/opt/csw/lib -liconv -R/opt/csw/lib   -lsocket -lnsl
libtool: link: /opt/solarisstudio12.4/bin/cc -G -z defs -h libpsl.so.5 -o .libs/libpsl.so.5.3.1  .libs/libpsl_la-psl.o .libs/libpsl_la-lookup_string_in_fixed_set.o   -R/opt/csw/lib -lidn2 -lunistring -L/opt/csw/lib -liconv -lsocket -lnsl -lc  -g  
ld: fatal: library -lidn2: not found
ld: fatal: library -lunistring: not found
gmake[3]: *** [Makefile:478: libpsl.la] Error 2
```

rockdaboot:
You forgot to set LD_LIBRARY_PATH:
```
export CC=/opt/solarisstudio12.4/bin/cc
export LDFLAGS=-L/opt/csw/lib
export LD_LIBRARY_PATH=/opt/csw/lib
./autogen.sh
./configure
make
make check
```

rockdaboot:
The bad news is that certain tests fail when built with cc. As if the section flags PRIVATE/ICANN are not properly detected. The generated data (DAFSA) files are bit identical, so it must be the libpsl code behaving differently with cc and gcc. Looks like a bug in cc. Any other explanation ?

dago:
Using `LD_LIBRARY_PATH` is discouraged as it changes linkage from invoked binaries during the build process. It can happen that different libraries with the same name exist in `/opt/csw/lib` and `/usr/lib` with the same name but different functionality and especially Sun binaries in `/usr/bin` must not use any library from `/opt/csw/lib` as this can result in undefined behaviour. The correct configuration is `LDFLAGS=-L/opt/csw/lib` and `LD_OPTIONS=-R/opt/csw/lib`. This is my main concern here.

The failed checks for Sun Studio can be verified after we fix the ignorance of `$(LDFLAGS)`.

rockdaboot:
I'm just testing on branch 'fix-build'. Please also give it a try. Currently only changed for libidn2 to see what Travis is saying.

dago:
Looks good. From what I see you reordered `libidn2` to come late on the command line, this works because the linker path has been included earlier already:
```
libtool: link: /opt/solarisstudio12.4/bin/cc -G -z defs -h libpsl.so.5 -o .libs/libpsl.so.5.3.1  .libs/libpsl_la-psl.o .libs/libpsl_la-lookup_string_in_fixed_set.o   -R/opt/csw/lib -L/opt/csw/lib -liconv -lunistring -lidn2 -lsocket -lnsl -lc  -g  
libtool: link: (cd ".libs" && rm -f "libpsl.so.5" && ln -s "libpsl.so.5.3.1" "libpsl.so.5")
libtool: link: (cd ".libs" && rm -f "libpsl.so" && ln -s "libpsl.so.5.3.1" "libpsl.so")
...
/bin/sh ../libtool  --tag=CC   --mode=link /opt/solarisstudio12.4/bin/cc  -g  -L/opt/csw/lib -o psl psl.o ../src/libpsl.la -lunistring -lidn2  -lsocket -lnsl
libtool: link: /opt/solarisstudio12.4/bin/cc -g -o .libs/psl psl.o  -L/opt/csw/lib ../src/.libs/libpsl.so -liconv -lunistring -lidn2 -lsocket -lnsl -R/usr/local/lib
gmake[2]: Leaving directory '/home/dam/work/libpsl/tools'
```

PR #113, opened 2018-12-06
WIP: Add initial Meson build system

xclaesse:
Still missing at least fuzz, doc, and man pages.

But it builds and tests are passing at least on my Ubuntu 18.04.

### DIFF
--- a/fuzz/meson.build
+++ b/fuzz/meson.build
@@ -1,0 +1,13 @@
+fuzzer_cargs = [
+  '-DSRCDIR="@0@"'.format(meson.source_root()),
+  '-DTEST_RUN',
+]
+
+foreach test_name : ['fuzzer', 'load_fuzzer', 'load_dafsa_fuzzer']
+  source_file = 'libpsl_@0@.c'.format(test_name)
+  exe = executable(test_name, source_file, 'main.c',
+    link_with : libpsl,
+    c_args : fuzzer_cargs,
+    include_directories : [configinc, includedir])
+  test(test_name, exe)
+endforeach
--- a/include/meson.build
+++ b/include/meson.build
@@ -1,0 +1,17 @@
+version_arr = meson.project_version().split('.')
+cdata = configuration_data()
+cdata.set('LIBPSL_VERSION', meson.project_version())
+cdata.set('LIBPSL_VERSION_MAJOR', version_arr[0])
+cdata.set('LIBPSL_VERSION_MINOR', version_arr[1])
+cdata.set('LIBPSL_VERSION_PATCH', version_arr[2])
+
+# FIXME: We should not run a command-line for this, but I don't think Meson
+# has API for that.
+printf = find_program('printf')
+r = run_command(printf, '0x%02x%02x%02x', version_arr, check : true)
+cdata.set('LIBPSL_VERSION_NUMBER', r.stdout().strip())
+
+configure_file(
+  input : 'libpsl.h.in',
+  output : 'libpsl.h',
+  configuration : cdata)
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,104 @@
+project('libpsl', 'c',
+  version : '0.20.2',
+  meson_version : '>=0.47.0')
+
+cc = meson.get_compiler('c')
+
+enable_runtime = get_option('runtime')
+enable_builtin = get_option('builtin')
+
+notfound = dependency('', required : false)
+libidn2_dep = notfound
+libicu_dep = notfound
+libidn_dep = notfound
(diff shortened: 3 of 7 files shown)


rockdaboot:
Great you work on this !

Please put the changes in `src/Makefile.am` and `src/psl.c` into an own commit (or even an own PR). Looks that I could merge it already.

rockdaboot:
And another PR for the changes in `tests/Makefile.am` and `tests/test-is-public-all.c`.

xclaesse:
I think for a first step it's already looking good. I can build libsoup with libpsl as subproject, and all libsoup unit tests pass. I realize this is useful on Ubuntu 18.04 because libsoup master depends on libpsl >= 0.20 but my distro has 0.19.

I think all the code and tests gets built now. Missing many other things, but maybe we could already merge this first step. What do you think?

rockdaboot:
Will be offline in a few seconds... so I'll take a look tomorrow. But in the end you decide - you are the expert and I likely won't use meson in the near future.

rockdaboot:
I just tried to build with meson, following some basic tutorial:
```
$ meson builddir
The Meson build system
Version: 0.48.2
Source dir: /home/oms/src/libpsl
Build dir: /home/oms/src/libpsl/builddir
Build type: native build
meson.build:18: WARNING: Identifier 'in' will become a reserved keyword in a future release. Please rename it.

meson.build:18:18: ERROR:  Expecting eol got id.
if enable_runtime in ['libidn2', 'auto'] or enable_builtin in ['libidn2', 'auto']
                  ^

A full log can be found at /home/oms/src/libpsl/builddir/meson-logs/meson-log.txt
```

xclaesse:
Oops, I'm using the 'in' keyword which is only supported on meson master. Will be released this weekend.

rockdaboot:
It might take while until that version is available as package for stable systems. Is there a more backwards compatible replacement for 'in' ?

xclaesse:
@rockdaboot sure, `foo in bar` is new syntax for `bar.contains(foo)`. I updated my PR now, I forgot that wasn't released yet. Note that Meson is still a fast moving target, distro version is often outdated. But you can easily install latest release with `pip3 install meson`.

Are you considering dropping the autotools build system when the meson one gets ready?

rockdaboot:
Thank you !

> Are you considering dropping the autotools build system when the meson one gets ready?

Not sorry, not a second. Libpsl builds (automatically) on many systems without meson, even on older systems without python3. Dropping it means breaking lot's of stuff. We also have a Microsoft 'nmake' build, as you have seen. I think all those might co-exist. I wouldn't even reject CMake build stuff.

But autotools is the officially supported build system. If ppl come up with questions regarding nmake or meson, I just can hope some else answers them or fixes stuff.

xclaesse:
> Not sorry, not a second. Libpsl builds (automatically) on many systems without meson, even on older systems without python3. Dropping it means breaking lot's of stuff. We also have a Microsoft 'nmake' build, as you have seen. I think all those might co-exist. I wouldn't even reject CMake build stuff.

Ok, given that you already have a python script in the build system, I didn't think depending on python3 would be a problem. Btw, meson supports build for MSVC (didn't try with libpsl) Windows users could be happy ;-)

So if you've got no objection, I think this is already a useful first step. More work is definitely needed to get on par with autotools, but I don't think we need to make everything in a single PR.

rockdaboot:
Just a quick response (I'll check closer this weekend):

Just `meson` for building all would be nice, as one would execute `make`. Is it possible to configure a default builddir, e.g. `builddir`?

Then, also from the project dir, `meson check` would be nice as replacement to `make check`.

Ok, maybe I should use `ninja` instead. But cd'ing into `builddir` and then calling whatever to build and test seems a tedious, then cd'ing back to projectdir to edit, back-and-forth. Is there a better way, I might just not understand the command sequences.

xclaesse:
```
meson builddir
ninja -C builddir
ninja -C builddir test
```

rockdaboot:
Another question. Where do I find the output of the tests after running `ninja -C builddir test` ? Those logs are important for my daily live as developer. This is not a blocker, just curiosity.

rockdaboot:
Thanks for your work ! I added the meson build procedure to README.md.

xclaesse:
> Another question. Where do I find the output of the tests after running `ninja -C builddir test` ? Those logs are important for my daily life as developer. This is not a blocker, just curiosity.

As far as I know all output should be in `builddir/meson-logs/testlog.txt`. You can also run tests with this command that prints output: `meson test -C builddir -v`. Is it what you were looking for?

PR #114, opened 2018-12-06
Prepare for meson

xclaesse:
- tests: Do not hardcode psl.dafsa and psl_ascii.dafsa file names

- Rename suffixes_dafsa.c to suffixes_dafsa.h
    
    It is a generated file that we #include, so it makes more sense to have
    .h extension. Also Meson tries to compile that file if it has .c
    extension.


rockdaboot:
Thank you !

Issue #115, opened 2018-12-08
Build failure with Meson


ebassi:
The Meson build with libICU enabled for runtime and built-in fails to build:

## Configuration

```
meson \
  --wrap-mode=nodownload \
  --prefix=/usr \
  --libdir=/usr/lib \
  --sysconfdir=/etc \
  --localstatedir=/var \
  --bindir=/usr/bin \
  --sbindir=/usr/bin \
  --datadir=/usr/share \
  --includedir=/usr/include \
  --libexecdir=/usr/libexec \
  --mandir=/usr/share/man \
  --infodir=/usr/share/info \
  -Dbuiltin=libicu \
  -Druntime=libicu \
  ..
```

## Build error

```
[20/28] Linking target fuzz/fuzzer.
FAILED: fuzz/fuzzer 
ccache cc  -o fuzz/fuzzer 'fuzz/fuzz@@fuzzer@exe/libpsl_fuzzer.c.o' 'fuzz/fuzz@@fuzzer@exe/main.c.o' -Wl,--no-undefined -Wl,--as-needed -O2 -g -m64 -mtune=generic -Wl,--start-group src/libpsl.so -Wl,--end-group '-Wl,-rpath,$ORIGIN/../src' -Wl,-rpath-link,/ostbuild/source/libpsl/_build/src  
/usr/lib/gcc/x86_64-gnomeostree-linux/5.3.0/../../../../x86_64-gnomeostree-linux/bin/ld: fuzz/fuzz@@fuzzer@exe/libpsl_fuzzer.c.o: undefined reference to symbol 'u_cleanup_56'
//lib/libicuuc.so.56: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

Trying on a more recent system, with a newer libICU, I get the same build error:

```
[23/28] Linking target fuzz/fuzzer.
FAILED: fuzz/fuzzer 
cc  -o fuzz/fuzzer 'fuzz/fuzz@@fuzzer@exe/libpsl_fuzzer.c.o' 'fuzz/fuzz@@fuzzer@exe/main.c.o' -L/home/ebassi/gnome/install/lib -Wl,--no-undefined -Wl,--as-needed -Wl,--start-group src/libpsl.so -Wl,--end-group '-Wl,-rpath,$ORIGIN/../src' -Wl,-rpath-link,/home/ebassi/gnome/source/libpsl/_build/src  
[...]

PR #116, opened 2018-12-08
Meson: Fix link of fuzzer tests

xclaesse:
libicu is an internal dependency of libpsl, so if an executable
uses libicu directly, linking on libpsl is not enough, it must also
link on libicu.

Closes: #115


rockdaboot:
This builds but the fuzzers are useless (do-nothing) without the data in fuzz/. So a `ninja test` reports ok, but in fact the fuzzers didn't test any corpora (data files).

xclaesse:
Looks like a bug in the test if it report OK without actually testing anything. Can you tell me how I'm supposed to pass that data?

rockdaboot:
Not a bug, it's on purpose because we don't include the fuzz corpora in the tarball.

If your could set -DSRCDIR=... like in fuzz/Makefile.am (`-DSRCDIR=\"$(abs_srcdir)\"`. The fuzzers construct the corpora directory name with SRCDIR+basename(argv[0])+".in" (see `fuzz/main.c`).

So libpsl_icu_fuzzer would look into `SRCDIR/libpsl_icu_fuzzer.in/`. SRCDIR being the absolute path of fuzz/ in the project directory.

PR #118, opened 2018-12-08
Meson: Fix srcdir pointing to root source dir instead of current subdir

xclaesse:
With that fuzz tests seems to be doing a lot more work.

### DIFF
--- a/fuzz/meson.build
+++ b/fuzz/meson.build
@@ -1,5 +1,5 @@
 fuzzer_cargs = [
-  '-DSRCDIR="@0@"'.format(meson.source_root()),
+  '-DSRCDIR="@0@"'.format(meson.current_source_dir()),
   '-DTEST_RUN',
 ]
 
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -9,7 +9,7 @@ psl_ascii_dafsa = custom_target('psl_ascii.dafsa',
   command : [python, psl_make_dafsa, '--output-format=binary', '--encoding=ascii', '@INPUT@', '@OUTPUT@'])
 
 tests_cargs = [
-  '-DSRCDIR="@0@"'.format(meson.source_root()),
+  '-DSRCDIR="@0@"'.format(meson.current_source_dir()),
   '-DPSL_FILE="@0@"'.format(psl_file),
   '-DPSL_TESTFILE="@0@"'.format(psl_test_file),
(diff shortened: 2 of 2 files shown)


rockdaboot:
Looks good :-) Thank you.

Next questions ;-)
- how to make an ASAN / UBSAN build ?
- how to run the tests using valgrind ?

Nothing urgent, but to use meson in a Travis CI runner, potentially replacing autotools at some point.

xclaesse:
>     * how to make an ASAN / UBSAN build ?

Meson has built-in support for that, you can configure with `meson -Db_sanitize=<value> builddir` where value is one of [none, address, thread, undefined, memory, address,undefined]. Does that cover all you need? To verify you can check the flags passed to the compiler by running `ninja -C builddir -v` to print command lines executed.

>     * how to run the tests using valgrind ?

Tests can be run with `meson test --wrap='valgrind --tool=memcheck'`. See http://mesonbuild.com/Unit-tests.html.

Is that enough for your needs?

rockdaboot:
Perfect, thanks for the info.

rockdaboot:
I fail to build with meson+ASAN+clang. Ubsan and valgrind seems to succeed, also everything succeeds when building with gcc.

If you could have a look into contrib/check-hard-meson if something obvious is missing !? It's a simple shell script, the ASAN part is commented out.

xclaesse:
I don't know anything about ASAN tbh, but I see Meson prints a warning:
```
WARNING: Trying to use undefined sanitizer on Clang with b_lundef.
This will probably not work.
Try setting b_lundef to false instead.
```

Here I get many linker failure liket this one:
```
/home/xclaesse/programmation/libsoup/subprojects/libpsl/builddir/../src/lookup_string_in_fixed_set.c:208: undefined reference to `__ubsan_handle_pointer_overflow'
```

xclaesse:
Note that with Meson we don't usually pass CFLAGS from the env. Better add them in the meson.build definition with `add_project_arguments(cc.get_supported_arguments([...]), language : 'c')`. Also the C std can be set in `project(default_options : ['c_std=gnu89'])`.

But that's details, it should work with CFLAGS too.

rockdaboot:
Looks like the asan build needs `-Db_sanitize=address -Db_lundef=false` *plus* some additional flags for the compiler (both clang and gcc): `-fsanitize=address -fno-omit-frame-pointer -fsanitize-address-use-after-scope`.

PR #120, opened 2018-12-11
Update the Meson build files so that things build on Windows

fanc999:
Hi,

As part of the steps to simplify the build system, this completes the support that is needed for libpsl to build on Windows using Meson, especially for Visual Studio builds.  After this gets merged, I think this will replace the NMake Makefiles that is used to build libpsl on Visual Studio.

With blessings, thank you!


rockdaboot:
Thanks !

PR #123, opened 2018-12-31
meson: install the library

nacho:
SSIA

### DIFF
--- a/include/meson.build
+++ b/include/meson.build
@@ -18,7 +18,9 @@ cdata.set('LIBPSL_VERSION_NUMBER',
     hex[minor / 16], hex[minor % 16],
     hex[patch / 16], hex[minor % 16]))
 
-configure_file(
+libpsl_h = configure_file(
   input : 'libpsl.h.in',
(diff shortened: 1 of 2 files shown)


xclaesse:
You probably also want to install headers, using `install_headers()` function. Are there other files that needs to be installed? I haven't looked at the part when doing the port, I'm using it uninstalled myself.

nacho:
@xclaesse I guess we need it installed for libsoup?

nacho:
@xclaesse done

xclaesse:
@nacho Now that I think about it, I'm surprised it works for me actually. Looks like that when I do `ninja install` in libsoup and libpsl is built as subproject, meson is smart enough to detect that libpsl.so is needed and it install it even without `install: true`. I just checked and when libpsl is built as standalone, libpsl.so is not installed, so this patch is needed.

xclaesse:
@nacho I haven't checked if autotools install other files as well, but this is at least a good start, LGTM.

rockdaboot:
merged manually

PR #124, opened 2019-01-09
Fix older windows

carlo-bramini:
At the moment the library requires Windows Vista or later, when working on Microsoft Windows.
I would like to suggest to not use inet_pton() on Windows and use the Winsock function that exists for the same purpose. The attached PR allows the library to run starting from Windows NT 2000. Actually the functions added for better compatibility with BSD are also based on Winsock, so I think that there are not pratical problems with this change.

Sincerely.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -310,8 +310,18 @@ elif test -n "$NEEDS_SOCKET" ; then
 elif test -n "$NEEDS_NSL" ; then
   LIBS="$LIBS -lnsl"
 else
-  # Needed for MinGW / Windows
-  AC_SEARCH_LIBS(inet_pton, ws2_32)
+  # Platform dependant options
+  case "${host_os}" in
+    # MinGW / Windows
+    *mingw*)
+      # Select Windows NT/2000 and later, for WSAStringToAddressW()
+      CPPFLAGS="$CPPFLAGS -D_WIN32_WINNT=0x500"
+      # Needed for network support
+      LIBS="$LIBS -lws2_32"
+      ;;
+    *)
+      ;;
+  esac
 fi
 
 # Check for clock_gettime() used for performance measurement
--- a/src/psl.c
+++ b/src/psl.c
@@ -48,9 +48,6 @@
 #include <sys/stat.h>
 
 #ifdef _WIN32
-/* This is for Windows Vista and later, for inet_pton() */
-# define _WIN32_WINNT 0x0600
-
 # include <winsock2.h>
 # include <ws2tcpip.h>
 #else
@@ -1578,10 +1575,26 @@ int psl_check_version_number(int version)
 /* return whether hostname is an IP address or not */
 static int isip(const char *hostname)
 {
+#ifdef _WIN32
+	WCHAR wName[INET6_ADDRSTRLEN+1];
+
+	struct sockaddr_in  addr  = {0};
+	struct sockaddr_in6 addr6 = {0};
+
+	INT size  = sizeof(addr);
+	INT size6 = sizeof(addr6);
+
+	if (!MultiByteToWideChar(CP_UTF8, 0, hostname, -1, wName, countof(wName)))
+		return 0;
+
+	return (WSAStringToAddressW(wName, AF_INET,  NULL, (struct sockaddr *)&addr,  &size) != SOCKET_ERROR) |
+	       (WSAStringToAddressW(wName, AF_INET6, NULL, (struct sockaddr *)&addr6, &size6) != SOCKET_ERROR);
+#else
 	struct in_addr addr;
 	struct in6_addr addr6;
 
 	return inet_pton(AF_INET, hostname, &addr) || inet_pton(AF_INET6, hostname, &addr6);
+#endif
 }
 
 /**


configure.ac
```diff
@@ -311,7 +311,7 @@ elif test -n "$NEEDS_NSL" ; then
   LIBS="$LIBS -lnsl"
 else
   # Needed for MinGW / Windows
-  AC_SEARCH_LIBS(inet_pton, ws2_32)
+  LIBS="$LIBS -lws2_32"
```
rockdaboot: This breaks the build outside Windows. The AC_SEARCH_LIBS is essential.

src/psl.c
```diff
@@ -1588,7 +1588,7 @@ static int isip(const char *hostname)
 	INT size  = sizeof(addr);
 	INT size6 = sizeof(addr6);
 
-	if (!MultiByteToWideChar(CP_UTF8, 0, hostname, -1, wName, sizeof(wName)/sizeof(WCHAR)))
+	if (!MultiByteToWideChar(CP_UTF8, 0, hostname, -1, wName, countof(wName)))
```
rockdaboot: please merge with the previous commit.

src/psl.c
```diff
@@ -49,6 +49,7 @@
 
 #ifdef _WIN32
 /* This is for Windows NT/2000 and later, for WSAStringToAddressW() */
+# undef _WIN32_WINNT
```
rockdaboot: This throws a warning with some compilers (e.g. clang) if _WIN32_WINNT is not defined. Since these are more often used than MSVC, please remove this commit.

src/psl.c
```diff
@@ -48,8 +48,8 @@
 #include <sys/stat.h>
 
 #ifdef _WIN32
-/* This is for Windows Vista and later, for inet_pton() */
-# define _WIN32_WINNT 0x0600
+/* This is for Windows NT/2000 and later, for WSAStringToAddressW() */
+# define _WIN32_WINNT 0x0500
```
rockdaboot: Please merge this commit with the next one, since both only work together.

rockdaboot:
As long as the patch doesn't break anything else, I don't reject it. Just because a vendor says something is EOL, that doesn't mean anything to OSS projects.

configure.ac
carlo-bramini: My fault, when I saw that comment, I had understood that it was just for that platform.
But I have to say that I'm not able to compile the original code with `AC_SEARCH_LIBS()`.
It always fails because it is trying to link with an `__stdcall` function.
This function requires that the right include files to have been included at compile time.
Otherwise it will be never found by the linker if to get the` __stdcall` qualification is not applied.
Both MSYS2/MINGW and CYGWIN/MINGW-W64 fail on my PC.
It is unclear to me how previous code could work because in the past I had seen that configure scripts tested `${host_os}` variable and added `-lws2_32` directly to bypass this issue.

src/psl.c
carlo-bramini: I added this line becase I'm getting this warning without it:

```
  CC       libpsl_la-psl.lo
../../src/psl.c:53:0: warning: "_WIN32_WINNT" redefined
 # define _WIN32_WINNT 0x0500

In file included from /usr/i686-w64-mingw32/sys-root/mingw/include/crtdefs.h:10:0,
                 from /usr/i686-w64-mingw32/sys-root/mingw/include/sys/types.h:13,
                 from ../../src/psl.c:47:
/usr/i686-w64-mingw32/sys-root/mingw/include/_mingw.h:232:0: note: this is the location of the previous definition
 #define _WIN32_WINNT 0x502
```

It is not a big problem however, if you prefer I can remove it.

configure.ac
rockdaboot: I do cross-compiling libpsl with MinGW without any issue, but don't test on Windows.
There is a Meson build arch, but I doubt that anyone tested it on WinXP.

Why not check `${host_os}` for windows and skip AC_SEARCH_LIBS in that branch ?

src/psl.c
carlo-bramini: Sorry, please excuse my ignorance... what do you mean with merge with previous commit?
Is it a thing that I can do with github web interface?

rockdaboot: I guess you can't use Github for that. You can do that by using 'git rebase -i' (you find examples on the web) and after that you force push, Github then amends your PR automatically.

configure.ac
carlo-bramini: I did the check on `${host_os}` in my branch and it seems to work fine.
BTW, do you think that is would be acceptable something like this:

```
else
  # Platform dependant options
  case "${host_os}" in
    # MinGW / Windows
    *mingw*)
      # Select Windows NT/2000 and later, for WSAStringToAddressW()
      CPPFLAGS="$CPPFLAGS -D_WIN32_WINNT=0x500"
      # Needed for network support
      LIBS="$LIBS -lws2_32"
      ;;
    *)
      ;;
  esac
fi
```

As you can see, I also moved the macro `_WIN32_WINNT` to `$CPPFLAGS` rather than having it declared inside the source (like VisualStudio does afterall) and it seems to have fixed both issues.

src/psl.c
rockdaboot: Just don't mind, I'll merge all your commits into one before adding your changes to master.

carlo-bramini:
Excuse me, I was wondering if the PR was looking good after my last code change.
Thank you very much.

rockdaboot:
Thanks for pinging me.

One question is why not to use `AC_SEARCH_LIBS(inet_pton, ws2_32)` ? Wouldn't that be a more general approach than checking for $host_os !?

Could you merge your commits into a single one, please ?

Since nobody complained, I assume the code around WSAStringToAddressW() is ok.

carlo-bramini:
Hello, thank you very much for your reply.

As I have written in my previous message, I was never able to build the source of libpsl on Windows as they are.
On Windows, a function can be declared for supporting different types of calling conventions and stack management.
The function `inet_pton()` is declared as `__stdcall` function, this means that the function automatically frees the arguments from the stack when it returns.
When a function is declared in this way, the compiler automatically appends a character "@" at the end of the function name, followed by the number of bytes consumed by the parameters. It also adds a prefix "_imp_" when the function must be imported or exported from a dinamic library, but this last one is handled transparently.
So, the symbol inside ws2_32 is not "_inet_pton", but it is _imp__inet_pton@12.
The problem comes during the configuration, because the configure script emits this test (I copied it directly from config.log):


```
configure:16991: checking for library containing inet_pton
configure:17022: i686-w64-mingw32-gcc -o conftest.exe -g -O2   conftest.c  >&5
/tmp/ccBp2MzP.o: In function `main':
/home/carlo/libpsl/a/conftest.c:32: undefined reference to `inet_pton'
collect2: error: ld returned 1 exit status
configure:17022: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "libpsl"
| #define PACKAGE_TARNAME "libpsl"
| #define PACKAGE_VERSION "0.20.2"
| #define PACKAGE_STRING "libpsl 0.20.2"
| #define PACKAGE_BUGREPORT "tim.ruehsen@gmx.de"
| #define PACKAGE_URL "https://github.com/rockdaboot/libpsl"
| #define STDC_HEADERS 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_MEMORY_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_UNISTD_H 1
| #define LT_OBJDIR ".libs/"
| #define HAVE_VISIBILITY 1
| /* end confdefs.h.  */
| 
| /* Override any GCC internal prototype to avoid an error.
|    Use char because int might match the return type of a GCC
|    builtin and then its argument prototype would still apply.  */
| #ifdef __cplusplus
| extern "C"
| #endif
| char inet_pton ();
| int
| main ()
| {
| return inet_pton ();
|   ;
|   return 0;
| }
configure:17022: i686-w64-mingw32-gcc -o conftest.exe -g -O2   conftest.c -lws2_32   >&5
/tmp/ccSLnmzg.o: In function `main':
/home/carlo/libpsl/a/conftest.c:32: undefined reference to `inet_pton'
collect2: error: ld returned 1 exit status

```

This failure happens because that test program declared "char inet_pton();" instead of "char __stdcall inet_pton ();".
And so, the "ws2_32" library is not added to LIBS and the compilation of the library fails with:


```
  CCLD     libpsl.la
.libs/libpsl_la-psl.o: In function `isip':
/home/carlo/libpsl/a/src/../../src/psl.c:1584: undefined reference to `_imp__inet_pton@12'
collect2: error: ld returned 1 exit status

```

The troubles with Windows and __stdcall/__cdecl/__fastcall calling conventions are known with autotools, that's why the preferred solution is to check if the target is Windows and just add the libraries if it is.
I have no idea how the code in the master branch could be compiled on Windows... Actually, I know that GNUC can add an alias to a function name, so "_imp__inet_pton@12" could also exist as "_imp__inet_pton" inside a library, but I tried to build with CYGWIN+MINGW-W64 and MSYS2+MINGW and both did not work. Actually, I have not other ideas to make it working here, but if you have some suggestion, I can do some tests.

BTW the reason because I suggested the usage of winsock API instead of `inet_pton()` was born because libpsl blocked the execution of KiCad EDA (http://kicad-pcb.org/) on Windows XP. At the moment, I compiled libpsl myself and I replaced the existing library with my compiled version, but if it could be possible to make it working again without much troubles, it would be great ;)

Sincerely.

rockdaboot:
Thanks for info and updating !

PR #126, opened 2019-04-04
Windows: MSVC doesn't define ssize_t

xhaakon:


rockdaboot:
Thanks ! Could you please also add that code to fuzz/main.c !?

And what do you think about
```
#if defined _MSC_VER && ! defined ssize_t
...
```
?

Just in case a newer version of MSVC has a ssize_t defined. Or is it too theoretical ?

xhaakon:
Hi,

> Thanks ! Could you please also add that code to fuzz/main.c !?

Added.

> ```
> #if defined _MSC_VER && ! defined ssize_t
> ```

Yes, that makes sense. Changed in both locations.

src/psl.c
```diff
@@ -56,6 +56,10 @@
 # include <unistd.h>
 #endif
 
+#if defined(_MSC_VER) && ! defined(ssize_t)
+typedef SSIZE_T ssize_t;
```
rockdaboot: What about basetsd.h here ? It might be included by winsock2.h, but we shouldn't rely on it...

xhaakon: Alright, added the include here as well.

rockdaboot:
Thanks for that !

PR #131, opened 2019-04-17
Fix build when configured with --with-psl-file

sunpoet:
The error message is as follows:
```
Making all in tests
make: don't know how to make ../list/public_suffix_list.dat. Stop

make: stopped in /usr/ports/works/usr/ports/dns/libpsl/work/libpsl-libpsl-0.21.0/tests
*** Error code 1

Stop.
```

PSL_FILE defaults to $(top_srcdir)/list/public_suffix_list.dat. When configured with --with-psl-file, the non-default PSL_FILE breaks the build in tests directory. Use $(PSL_FILE) instead of hardcoded $(top_srcdir)/list/public_suffix_list.dat to fix the build.


rockdaboot:
Thank you !

PR #137, opened 2019-06-01
gtk-doc: do not include tree_index.sgml

kanavin:
gtk-doc 1.30 no longer generates the file if the object tree is empty.


rockdaboot:
Thank you !

PR #140, opened 2019-07-11
Makefile.am: use PYTHON when invoking psl-make-dafsa

rossburton:
In an environment where only Python 3 is installed, configure.ac finds and sets
PYTHON=python3 correctly but src/psl-make-dafsa is called directly, so the hashbang of
`#!/usr/bin/env python` is used which doesn't exist.

Fix this by explicitly running $(PYTHON) when using the tool.


rossburton:
Note that I haven't tested the same environment with Meson so that may have the same problem.

rockdaboot:
Thanks, Ross. That's a good idea. (I personally don't use meson, but any complaints from those users will be addressed at that time).

Issue #145, opened 2019-10-29
Missing build instructions on Windows


dukavi:
There is msvc folder with makefiles and some scripts, but no instructions how to use it.

rockdaboot:
@fanc999 Could you possibly help out !?

rockdaboot:
#146 closes this, thank you @fanc999

PR #146, opened 2019-10-30
Improve Visual Studio builds Add instructions

fanc999:
Hi,

This attempts to improve Visual Studio NMake builds by:

*  Fix test program builds, as new macro defines are needed to locate the generated dafsa data files.
*  Support Visual Studio 2019 better.
*  Separate intermediate PDBs during the build.
*  Differentiate builds between Visual Studio versions.

Since we also support Meson for Visual Studio builds, we also:
*  Add instructions on building libpsl on Visual Studio with NMake and Visual Studio (issue #145).

With blessings, thank you!


rockdaboot:
That looks great, thank you so much.

PR #148, opened 2020-02-28
meson: Use libtool library versioning, same as autotools

nirbheek:
This helps maintain ABI compatibility with the Autotools build so it's a drop-in replacement.

Previously we only outputted `libpsl.so`. Now we output:

```
libpsl.so -> libpsl.so.5
libpsl.so.5 -> libpsl.so.5.3.2
libpsl.so.5.3.2
```

Which is ABI-compatible with what autotools outputs:

```
libpsl.so -> libpsl.so.5.3.2
libpsl.so.5 -> libpsl.so.5.3.2
libpsl.so.5.3.2
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2,6 +2,9 @@ project('libpsl', 'c',
   version : '0.20.2',
   meson_version : '>=0.47.0')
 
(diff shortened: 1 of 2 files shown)


rockdaboot:
Thank you !

PR #149, opened 2020-03-10
Fix windows msvc issues

nirbheek:
ssize_t is never defined as a macro on MSVC

strings.h is not available when building with MSVC

Noticed at: https://gitlab.freedesktop.org/gstreamer/gst-build/issues/84

### DIFF
--- a/fuzz/main.c
+++ b/fuzz/main.c
@@ -41,7 +41,7 @@
 #include <errno.h>
 #include <sys/stat.h>
 
-#if defined (_MSC_VER) && ! defined (ssize_t)
+#if defined(_MSC_VER)
 #include <basetsd.h>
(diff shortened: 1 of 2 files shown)


rockdaboot:
Thanks for your contribution.

I just added a configure (and meson) check for strings.h. That seems a bit more general then checking for _MSC_VER. Can you test if this works for you ?

The second thing is the removal of the compile time check for ssize_t. AFAIR, this was for a reason. So if it doesn't break your compilation, I would like to leave it in for now.

rockdaboot:
Right, some devs like to define `ssize_t` on their own on the command line (or via IDE). That's why there is `... && ! defined(ssize_t)`.

nirbheek:
> Can you test if this works for you ?

This works, thanks!

PR #150, opened 2020-04-02
meson: run python script against the meson version of python

ystreet:
Avoids using a python that may not exist in a python3 only distribution:

/usr/bin/env: ‘python’: No such file or directory

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -3,7 +3,7 @@ psl_make_dafsa = find_program('psl-make-dafsa')
 suffixes_dafsa_h = custom_target('suffixes_dafsa.h',
   input : psl_file,
   output : 'suffixes_dafsa.h',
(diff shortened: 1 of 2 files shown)


rockdaboot:
` ‘python’: No such file or directory` means there is no 'python' command in $PATH. As I understand, your change also tries to run a 'python' command. Please explain why or how that works when 'python' isn't in $PATH.

ystreet:
The python command is retrieved from meson which is python3 generally as meson is written using python3.  On systems without python2 (fedora), the symlink between python -> python2 does not exist.

This is also the same thing for meson as was merged for autotools in https://github.com/rockdaboot/libpsl/commit/b4fec5d0ddb70fc4f5360eb14f2f5c5e91194333

rockdaboot:
Ah, ok. Thanks for your explanation.

PR #151, opened 2020-04-03
meson: Use the b_vscrt option for selecting the CRT

nirbheek:
This option has been available since 0.48, and we should use it
instead of only guessing based on buildtype.


rockdaboot:
I rely on your expertise :-)

Issue #136, opened 2019-05-26
0.21.0: meson does not install gtk-doc and man3 pages


kloczek:
Looks like that bits are missing still in meson :)

rockdaboot:
@xclaesse, @nacho Could you take a look ?

eli-schwartz:
Partially addressed by #168, fully addressed by #219

PR #163, opened 2020-10-25
Add 'Building from tarball' to README.md

rockdaboot:


meson.build
```diff
@@ -10,24 +10,15 @@ cc = meson.get_compiler('c')
...
+# We need to know the build type to determine what .lib files we need on Visual Studio
+# for dependencies that don't normally come with pkg-config files for Visual Studio builds
+buildtype = get_option('buildtype')
 
 notfound = dependency('', required : false)
 libidn2_dep = notfound
 libicu_dep = notfound
 libidn_dep = notfound
 libunistring = notfound
-libiconv = notfound
```
tp-m, line 30: Is this being removed again on purpose? This was just added in a previous PR #136 

Or was it removed by accident during conflict resolution?

meson.build
```diff
@@ -95,7 +86,6 @@ endif
 if libidn2_dep.found() or libidn_dep.found()
   # Check for libunistring, we need it for psl_str_to_utf8lower()
   libunistring = cc.find_library('unistring')
-  libiconv = cc.find_library('libiconv')
```
tp-m, line 98: Is this being removed again on purpose? This was just added in a previous PR #136 

Or was it removed by accident during conflict resolution?

tp-m (commented):
Some meson bits got removed here which seems to break the build: 
```
subprojects\libpsl\src\meson.build:19:0: ERROR: Unknown variable "libiconv".
```

meson.build
rockdaboot: The meson.build seems to be auto-generated from time to time, not sure if by 'make' or './configure'. Unnoticed by me this time. Sorry for that, but if that ever happens again, I have to remove `meson.build` from the repo (auto-generated files don't belong there).

meson.build
rockdaboot: See comment below.

PR #164, opened 2020-10-26
meson: fix build

tp-m:
src/meson.build:19:0: ERROR: Unknown variable "libiconv".

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -19,7 +19,7 @@ cargs = [
 libpsl = library('psl', sources, suffixes_dafsa_h,
   include_directories : [configinc, includedir],
   c_args : cargs,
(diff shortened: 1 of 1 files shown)


tp-m:
Also see https://github.com/rockdaboot/libpsl/pull/163#pullrequestreview-516589282

tp-m:
Would be great if we could get this merged as a matter of urgency, it's currently breaking all kinds of CI flows that foolishly are using libpsl master as suproject instead of pinning it to a release :)

nirbheek (approved):
LGTM

rockdaboot:
The meson stuff isn't tested by me as I don't use it. So currently it's up to you meson experts to keep it working :smile:.

I'd highly appreciate if someone of you come up with CI meson build, either on Travis or Github actions (or whatever you prefer).
Just to make merges a bit safer.

PR #166, opened 2021-01-07
Add libiconv dep to meson build again

quink-black:
Build success on macOS and ubuntu.


rockdaboot:
Thanks, LGTM. (Works for me on Debian unstable.)

Issue #156, opened 2020-07-18
0.21.1: meson does not install psl binary and man page


kloczek:
Looks like meson does not install psl executable and its man page.
```
+ /usr/bin/meson install -C x86_64-redhat-linux-gnu --no-rebuild
Installing src/libpsl.so.5.3.2 to /home/tkloczko/rpmbuild/BUILDROOT/libpsl-0.21.1-2.fc33.x86_64/usr/lib64
Installing /home/tkloczko/rpmbuild/BUILD/libpsl-0.21.1/x86_64-redhat-linux-gnu/include/libpsl.h to /home/tkloczko/rpmbuild/BUILDROOT/libpsl-0.21.1-2.fc33.x86_64/usr/include/
Installing /home/tkloczko/rpmbuild/BUILD/libpsl-0.21.1/x86_64-redhat-linux-gnu/meson-private/libpsl.pc to /home/tkloczko/rpmbuild/BUILDROOT/libpsl-0.21.1-2.fc33.x86_64/usr/lib64/pkgconfig
```

kloczek:
Also it does not allow build and install gtk-doc documentation.

PR #167, opened 2021-01-15
Meson: install libpsl man page, DAFSA builder

SteelPangolin:
Partially addresses issue #136, at least when building from a distribution tarball.

Note: Homebrew builds `libpsl` from a tarball using Meson:
https://github.com/Homebrew/homebrew-core/blob/0d0eb30/Formula/libpsl.rb

Thus, macOS users using Homebrew don't get man pages or the DAFSA builder. As of now, Homebrew is building from the 0.21.1 release, which doesn't have the fix for #156, so the `psl` utility isn't available on macOS either. If this is accepted, I suggest cutting a 0.21.2 release including this fix and #158 so that Homebrew can have a new release tarball to build from. I'm happy to submit the Homebrew formula patch.


rockdaboot:
Hey, thanks, LGTM.

docs/libpsl/meson.build
```diff
@@ -0,0 +1 @@
+install_man('libpsl.3')
```
nacho, line 1: this is wrong no? this file does not exist

SteelPangolin:
It's not in the source tree, but it does get built by the automake build
and included in release tarballs. It works if you build with Meson from a
tarball, but I think you're correct that this would break building with
Meson directly from a Git checkout. Basically, issue #136 is still an issue.

PR #168, opened 2021-01-26
meson: build reference manual with gtk_doc

nacho:


rockdaboot:
No complaints in 5 days - seems to be good :-) (I don't use meson and there are no automated checks for a meson build.)

rockdaboot:
TY !

PR #178, opened 2022-01-04
Fix Meson build

equeim:
Remove meson.build generation and store project/library versions in separate files which are read by autoconf/Meson.
Bring back libiconv dependency lookup (looks like it was removed accidentally).
Remove libicu includes test-registrable-domain.c.


meson.build
```diff
@@ -86,6 +87,19 @@ endif
 if libidn2_dep.found() or libidn_dep.found()
   # Check for libunistring, we need it for psl_str_to_utf8lower()
   libunistring = cc.find_library('unistring')
+  found_iconv = false
+  if cc.has_function('iconv_open')
+    libiconv_dep = []
+    found_iconv = true
+  endif
+  if not found_iconv and cc.has_header_symbol('iconv.h', 'iconv_open')
```
eli-schwartz, line 95: Note that once you are able to upgrade your minimum meson requirement to 0.60, you can replace these 13 lines with just `dependency('iconv')`.

rockdaboot:
@equeim `meson.build` is auto-generated from `meson.build.in`. Please make your changes in that file, else they get lost. Same for #177.

Since I don't work with meson, maybe you can check whether this auto-generation (made by `./configure`) is still needed. And if so, `meson.build` should be removed from the git repository (please feel free to just add another commit for the removal to this PR).

equeim:
It seems like meson.build generation is used to synchronize project version between autotools and meson. I your are ok with raising meson version requirement to 0.57 then it would be possible to do it without generation (storing version in a separate file that autotools and meson would read from).
BTW, library version is still not synchronized (this can be done in the same way).

rockdaboot:
I don't mind raising to whatever meson version. I'd possibly wait a week before merging to give other meson users time to complain =)
Ideally, the meson code takes the version directly from configure.ac. As autotools is the primary build system. Do you think that is doable ?

equeim:
In case of project version (not so version), there are only two ways to do it with meson without file generation: either hardcode it in meson.build or place it in separate file (so that file only contains version string) and meson will read it. In second case there is no way to do any custom parsing logic, meson won't allow it (it is possible with so version though), so we can't extract it from configure.ac. I'm assuming that configure.ac will be able to read the same file since it will be executed as shell script after processing, so it will still come from single source.

eli-schwartz:
Seems like this was originally the fault of #166 which added this, but not to the template. :/

Regarding the version file, meson expects a file with a single line, and the entire contents of the file are treated as a version string. In shell script based build systems, the version would be `$(cat ./VERSION)`. :)

Since this is autotools, the build system is not (at that point, anyway -- AC_INIT is part of configure.ac, not configure!) a shell script based build system -- it is an m4-and-shell-and-perl-based build system. In that case, use `m4_normalize(m4_include([VERSION]))`.

This will even take care of adding the file to EXTRA_DIST, and triggering aclocal to run again and regenerate configure.ac -> configure when that file is modified.

rockdaboot:
Thanks, looks good to me :-)

Tested locally building with autotools (from git and from tarball) as well as bulding with meson from tarball.

I like the idea of having shared version files.

PR #179, opened 2022-01-07
Use C++ linker when building with libicu

equeim:
ICU is a C++ library, event though it has C interface. This means that it depends on C++ standard library internally. When ICU is built as dynamic library we can link to it using C linker without issues because it is already linked to C+ standard library (I'm not sure how portable is this but it works). However, if ICU is built as static library, we must link to C++ standard library manually or use C++ linker, or else we will get linking errors.

This commit does this for Meson. I'm not sure how to do this properly for autotools (I was able to find [this](https://www.gnu.org/software/automake/manual/automake.html#How-the-Linker-is-Chosen) in the automake manual though).


rockdaboot:
Thank you !

PR #182, opened 2022-05-30
Make the generated public suffix list constexpr

wangwenx190:


wangwenx190:
ping?

rockdaboot:
@wangwenx190 Sorry for the delay, I somehow missed this PR.

`constexpr` is C++, while this tool is needed to create C compatible code for libpsl. I see the comment above is misleading. What about changing it from `C++` to `C` ?

wangwenx190:
> @wangwenx190 Sorry for the delay, I somehow missed this PR.
> 
> `constexpr` is C++, while this tool is needed to create C compatible code for libpsl. I see the comment above is misleading. What about changing it from `C++` to `C` ?

Can I read it as "the comments are wrong and need to be corrected, this PR is no more needed"?

rockdaboot:
> > @wangwenx190 Sorry for the delay, I somehow missed this PR.
> > `constexpr` is C++, while this tool is needed to create C compatible code for libpsl. I see the comment above is misleading. What about changing it from `C++` to `C` ?
> 
> Can I read it as "the comments are wrong and need to be corrected, this PR is no more needed"?

Yes.
The library code is C89.

wangwenx190:
OK, thanks for your reply! I'll close this PR then.

rockdaboot:
Thanks for bringign this up and sorry for your spent time.
I just pushed a fix for the comments (using C/C++ instead of C++).

wangwenx190:
It doesn't matter, no need to apologize. :)

PR #183, opened 2022-06-10
meson: switch iconv to meson's handling

neheb:
Fixes compilation under OpenWrt.

https://github.com/openwrt/packages/actions/runs/2471637273

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project('libpsl', 'c',
   version : files('version.txt'),
-  meson_version : '>=0.57.0')
+  meson_version : '>=0.60.0')
 
 cc = meson.get_compiler('c')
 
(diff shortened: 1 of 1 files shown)


eli-schwartz:
I previously mentioned this at https://github.com/rockdaboot/libpsl/pull/178#discussion_r778427646 but at the time I assumed/hoped it would just be a cleanup and thus not worth bumping the minimum version of Meson just for that.

It turns out that the builtin handling also fixes outright buggy misdetection on systems where both a builtin iconv and an external library exist. :D Well worth bumping the minimum IMO.

neheb:
Successful run with basically this patch: https://github.com/openwrt/packages/actions/runs/2471806701

rockdaboot:
Thanks for your contribution :+1:

PR #186, opened 2022-07-26
meson: fix regression that made it impossible to build

### LINKED ISSUE #185: operator of bool does not accept objects of type str
hi,

am I missing some dependency to build libpsl on Centos?

```
Compiler stderr:
Run-time dependency libidn2 found: YES 2.3.1
../meson.build:37:7: ERROR: The `==` operator of bool does not accept objects of type str (auto)
```

Thanks,
Dave

eli-schwartz:
In commit aa4909766c24c17bd1d9000ca419f6dc6b32c238 an option was modified and changed its type from a list of possible strings to a true/false value.

However, the build wasn't updated to correspond to that change, so it didn't logically work -- and Meson even threw a type error saying you cannot compare strings to booleans.

The actual change was basically "do not use this option anymore except to pass a macro to the source file and embed the data at all". So we can simply get rid of a bunch of complex dependency conditionals.

Fixes #185

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -34,10 +34,7 @@ if ['libidn2', 'auto'].contains(enable_runtime)
     if enable_runtime == 'auto'
       enable_runtime = 'libidn2'
     endif
-    if enable_builtin == 'auto'
-      enable_builtin = 'libidn2'
-    endif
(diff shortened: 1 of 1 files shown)


rockdaboot:
Thanks for taking care !

PR #187, opened 2022-10-11
Fix "Although the value stored to 'pos' is used in the enclosing expression, the value is never actually read from 'pos'"

Coeur:
Xcode analyser warning.
Capture d’écran 2022-10-11 à 17 46 40


rockdaboot:
Nice find, thanks !

PR #189, opened 2022-11-06
Use AC_SYS_LARGEFILE

mattst88:
Ensures that calls to fopen() and stat() can handle largefiles.


rockdaboot:
Huh. Now you made me curious what use case requires this 😃 
(Not saying I won't merge this PR)

mattst88:
Probably nothing, strictly speaking :)

But ChromeOS requires that largefile support is enabled for all packages—which is how I noticed this when trying to package it for ChromeOS. You can probably imagine how much of a pain it could be if something rarely failed one one user's system somewhere because a file was > 2GiB :)

rockdaboot:
> ChromeOS requires that largefile support is enabled for all packages
Thanks, TIL

PR #195, opened 2022-12-27
tools/psl.c: Fix build on Windows

fanc999:
Hi,

From the commit message:

<i>localtime_r() is not available on Windows but a more-secure variant of localtime(), localtime_s() is provided on Windows.</i>

<i>Define localtime_r() on Windows as its arguments are reversed as compared to localetime_s(), to achive more or less the same purpose.</i>

With blessings, thank you!

### DIFF
--- a/tools/psl.c
+++ b/tools/psl.c
@@ -38,6 +38,10 @@
 
 #ifdef _WIN32
 # include <winsock2.h> // WSAStartup, WSACleanup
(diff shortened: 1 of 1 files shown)


danyeaw:
Thanks @fanc999! I just ran in to this issue today and I was happy to see that it was already fixed. 👍

PR #200, opened 2023-01-14
add 'tests' option to disable tests and fuzzers

vtorri:


rockdaboot:
@eli-schwartz or @fanc999, could one of you please review/test this ?

rockdaboot:
@vtorri I would like to avoid breaking existing build scripts. Would it be possible to have a `no-tests` or `skip-tests` option to skip the tests. And "normal" operation stays as is ?

vtorri:
@rockdaboot done

eli-schwartz:
Really, I think the best approach as a matter of general principle is to always define the targets, but mark them as `build_by_default: false` so that they are not build when running `ninja`.

There's a small snag here, in that test dependencies are also built by default, even if they otherwise would not be -- this is still on my list of things to fix, they should only be built as a prerequisite when running `ninja test`

rockdaboot:
@eli-schwartz Since meson isn't my construction site, I'd like you to decide / approve (added you as collaborator, please feel free to say no).

eli-schwartz:
Sure! :)

meson.build
```diff
@@ -132,6 +132,8 @@ endif
 subdir('include')
 subdir('src')
 subdir('tools')
-subdir('tests')
-subdir('fuzz')
+if get_option('skip-tests') == false
```
eli-schwartz, line 135:
```suggestion
if not get_option('skip-tests')
```

However, as I notice the original commit actually made the tests option be true by default -- it would run the tests by default, which is behavior preserving, and to disable the tests you actually have to run with `-Dtests=false`

- `-Dtests=false`
- `-Dskip-tests=true`

Semantically, these mean the same thing, and I'm inclined to think the first one reads a bit more naturally. So I'd probably just drop the second commit in the end.

eli-schwartz:
> Really, I think the best approach as a matter of general principle is to always define the targets, but mark them as `build_by_default: false` so that they are not build when running `ninja`.

I have tweaked the PR to do this as well.

Now there is both:
- an option to disable tests (tests are built by default)
- even when tests are enabled, they are only part of the `test` target, not the `all` target -- so current versions of Meson will build them by default anyway, but future versions of Meson will (hopefully) not do so, until you actually run the `test` target.

rockdaboot:
Thanks for your contribution @vtorri and @eli-schwartz 😃

eli-schwartz:
My pleasure. :) I'm more than happy to handle any Meson PRs, so just point them my way. :)

eli-schwartz:
> even when tests are enabled, they are only part of the `test` target, not the `all` target -- so current versions of Meson will build them by default anyway, but future versions of Meson will (hopefully) not do so, until you actually run the `test` target.

Incidentally, I did finally get around to fixing this in meson upstream: https://mesonbuild.com/Release-notes-for-1-7-0.html#test-targets-no-longer-built-by-default

The test option is still useful for people on older versions of meson, but upgrading to meson 1.7.0 automatically reaps the benefit now.

PR #203, opened 2023-01-16
Improve NMake Makefiles for ARM64 builds and Visual Studio 2022

fanc999:
Hi,

Although we are supporting builds on Visual Studio via <strike>NMake</strike>Meson, I think it is a good idea to improve the NMake Makefiles as we still support that, by:

* Supporting ARM64 builds, with Visual Studio 2017 15.9.x or later.
* Distinguishing builds with Visual Studio 2019 and 2022.

With blessings, thank you!


rockdaboot (approved):
:+1:

PR #206, opened 2023-02-13
meson: add support for symbol visibility

eli-schwartz:
libpsl.h can set up a PSL_API macro handling cross-platform symbol hiding/exporting, which depends on indicating that:

- we're currently building the library
- the compiler supports visibility attributes

This avoids leaking symbols and lets the compiler optimize calls.


rockdaboot:
Thanks !

Issue #205, opened 2023-02-13
0.21.2 exports two symbols without `psl_` prefix


dkg:
I'm trying to get 0.21.2 in shape for debian packaging, and i notice that the shared object from 0.21.2 exports two symbols that  which don't have a `psl_` prefix:

- `GetUtfMode`
- `LookupStringInFixedSet`

I'm assuming that these are only accidentally exposed.  I haven't been able to trace why they're being exposed now, but they were not exposed in 0.21.1.

dkg:
i notice a6397a0d62992d5e73be9071c86a746c32fdac4e, which looks relevant, but it happened before 0.21.1 (before 0.20.2, actually) 

So it's possible that this is due to some sort of toolchain issue, i'm not sure.

I do notice that if i try to build from the git tags directly with `./autogen.sh && ./configure && make` that  i cannot deliberately replicate the problem.  So maybe there are some options that debian build infrastructure are using that make it possible to accidentally export these symbols?

eli-schwartz:
It seems it is conditional on gnulib providing `gl_VISIBILITY` which causes configure to test whether -fvisibility=hidden works, and if so, provide that as CFLAGS_VISIBILITY. Building from a freshly autoreconf'ed source tree I see this in the ./configure output:

```
./configure: 13558: gl_VISIBILITY: not found
```

eli-schwartz:
#206 solves this for meson, as it seems to be an oversight that it is missing there.

dkg:
i agree with this assessment, @eli-schwartz  -- thanks for looking into it so promptly!  it looks to  me like the gnulib macro `m4/visibility.m4` isn't included in the git repository, which is why `gl_VISIBILITY` is missing.  I'll prepare a patch for that, since i believe the expectation for gnulib is that it is copied into projects where it is used.

eli-schwartz:
That was my assumption too -- I had to download a copy from gnulib.git with `wget` and copy it into place, after which it worked, but I have no clue how gnulib works in general.

PR #207, opened 2023-02-13
Add m4/visibility.m4 from gnulib 20230113+stable-1 (Closes: #205)

dkg:
Without this, the gl_VISIBILITY macro in configure.ac doesn't work, which leaves some extra symbols exposed in some cases when building with autotools.

See https://github.com/rockdaboot/libpsl/issues/205


rockdaboot:
Thanks, good finding !.

PR #209, opened 2023-03-22
Fix compilation with libidn2 runtime on Windows

equeim:


neheb:
how is libidn2 added? vcpkg? I don't see this issue under MSYS2:

https://github.com/neheb/wrapdb/pull/4

rockdaboot:
> how is libidn2 added? vcpkg? I don't see this issue under MSYS2:
> 
> [neheb/wrapdb#4](https://github.com/neheb/wrapdb/pull/4)

We possibly need something slightly amended, like the first answer in https://stackoverflow.com/questions/3694723/error-c3861-strcasecmp-identifier-not-found-in-visual-studio-2008

rockdaboot:
Thanks for your contribution.
It made me take another look at the use of strcasecmp. And it was just used to check if a string is not `utf-8`. I just pushed a hand-rolled version of this test to increase portability.

PR #210, opened 2023-07-12
meson: Do not use full_path for file objects

nirbheek:
The full path will be inserted automatically here.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -13,8 +13,8 @@ tests_cargs = [
   '-DSRCDIR="@0@"'.format(meson.current_source_dir()),
   '-DPSL_FILE="@0@"'.format(psl_file),
   '-DPSL_TESTFILE="@0@"'.format(psl_test_file),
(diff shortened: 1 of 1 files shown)


eli-schwartz:
> The full path will be inserted automatically here.

Unfortunately, this is false. It would work *if* the custom_target was added as its own array element to a `command:` kwarg. The generated compiler macro is:
```
'-DPSL_DAFSA="<CustomTarget 59830eb@@psl.dafsa@cus: ['"'"'/usr/bin/python3.10'"'"', <File: src/psl-make-dafsa (not built)>, '"'"'--output-format=binary'"'"', '"'"'@INPUT@'"'"', '"'"'@OUTPUT@'"'"']>"'
```

https://github.com/mesonbuild/meson/pull/11834 would make this into a warning:

```
NOTICE: Future-deprecated features used:
 * 1.2.0: {'str.format: Value other than strings, integers, bools, options, dictionaries and lists thereof.'}
```

eli-schwartz:
Yup, that looks a lot safer and should probably work.

Issue #216, opened 2023-10-18
Doesn't build with clang-cl: undeclared library function `alloca`


hrxi:
```
../subprojects/libpsl-0.21.2/tests/test-is-public-all.c(121,19): error: call to undeclared library function 'alloca' with type 'void *(unsigned long long)'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
                char *xdomain = alloca(len + 3);
                                ^
../subprojects/libpsl-0.21.2/tests/test-is-public-all.c(121,19): note: include the header <stdlib.h> or explicitly provide a declaration for 'alloca'
```

Seen in the PR https://github.com/mesonbuild/wrapdb/pull/1260.

neheb:
According to ChatGPT, malloc.h must be included for alloca under Windows. Why is this using alloca anyway and not VLAs?

hrxi:
According to the error message by clang-cl (see initial post), it's `#include <stdlib.h>`.

rockdaboot:
> According to ChatGPT, malloc.h must be included for alloca under Windows. Why is this using alloca anyway and not VLAs?

One reason is that MSVC doesn't support VLAs at all.

rockdaboot:
`tests/test-is-public-all.c` is possibly not the only file to fix.

I won't have much time in the next days, so feel free to create a PR with something like
```
diff --git a/tests/test-is-public-all.c b/tests/test-is-public-all.c
index baee899..26b3c8a 100644
--- a/tests/test-is-public-all.c
+++ b/tests/test-is-public-all.c
@@ -38,6 +38,8 @@
 #include <ctype.h>
 #ifdef HAVE_ALLOCA_H
 #      include <alloca.h>
+#elif defined _WIN32
+#      include <malloc.h>
 #endif
 
 #include <libpsl.h>
```
Wherever it needs to be.

Reference: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/alloca?view=msvc-170

PR #217, opened 2023-11-02
fix compilation with clang-cl

neheb:
clang-cl errors on implicit functions. Needs the proper header included.

Fixes https://github.com/rockdaboot/libpsl/issues/216


rockdaboot (approved):
:+1:

PR #231, opened 2024-02-04
Avoid alloca in libpsl and tests

rockdaboot:
`alloca()` needs special treatment for Windows and possibly other systems. The implementation also is compiler-specific. `alloca()` in general is considered "dangerous" (similar to variable-length arrays (VLAs)).

This PR drops the usage of `alloca()` for the aforementioned reasons.

### DIFF
--- a/src/psl.c
+++ b/src/psl.c
@@ -69,12 +69,8 @@ typedef SSIZE_T ssize_t;
 
 #ifndef _WIN32
 # include <arpa/inet.h>
-#endif
-
-#ifdef HAVE_ALLOCA_H
-#	include <alloca.h>
-#elif defined _WIN32
-#	include <malloc.h>
+#else
+# include <malloc.h>
 #endif
 
 #ifdef WITH_LIBICU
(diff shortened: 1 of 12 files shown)


src/psl.c
```diff
@@ -78,12 +78,7 @@ typedef SSIZE_T ssize_t;
 
 #ifndef _WIN32
 # include <arpa/inet.h>
-#endif
-
-#ifdef HAVE_ALLOCA_H
-#	include <alloca.h>
-#elif defined _WIN32
-#	include <malloc.h>
+# include <malloc.h>
```
eli-schwartz, line 81: I assume you mean stdlib.h here? This is a glibc-ism.

eli-schwartz (changes requested):

src/psl.c
rockdaboot: Regarding https://learn.microsoft.com/de-de/cpp/c-runtime-library/reference/malloc?view=msvc-170, `malloc()` requires `stdlib.h` and `malloc.h`. The first is included further up, unconditionally.

Glibc only requires `stdlib.h`.

eli-schwartz: Right but for non Windows the canonical definition is stdlib.h -- malloc.h may not exist and isn't documented as the correct way to do it.

It does exist on glibc but it doesn't exist on macOS, which caused a build failure.

rockdaboot: My bad, the include was in the `#ifndef _WIN32` path, should have be in the `#else` path. Fixed it and force-pushed.

rockdaboot: @eli-schwartz Can you check again (see my comment https://github.com/rockdaboot/libpsl/pull/231#discussion_r1481061750) and unblock the PR if OK?

eli-schwartz: Yup, sorry, lost track of that.

eli-schwartz:
Rebased in order to refresh CI checks, everything looks good.

PR #233, opened 2024-02-05
CI: Add Github Actions testing for the meson builds too

eli-schwartz:


eli-schwartz:
For the meson builds I didn't bother to check all the different runtimes, in part because doing this nicely in a matrix is hard.

macOS finds libidn2 and Ubuntu finds ICU, though.

rockdaboot:
:+1:
Just learned that a) meson + ninja can be installed via pip and b) windows and macos runners exist for GH actions 😃

eli-schwartz:
Meson is supposed to be because it's python software. Ninja technically should be installed via  apt or brew but this introduced more complexity than I wanted compared to using a GH Actions matrix. It is handy, yeah.

PR #240, opened 2024-03-24
Add configure flag --with-libunistring-prefix

### LINKED ISSUE #214: Direct dependency "libunistring" has no configure options
When building libpsl with `--enable-runtime=libidn2` **libunistring** also seems to become a *direct* dependency. While libidn2 can be tuned via pkg-config or the LIBIDN2_CFLAGS/ LIBIDN2_LIBS variables there's no such visible option for libunistring in libpsl's `configure --help` output. The libunistring package doesn't offer any pkg-config integration too sadly… resulting in the libpsl configure just fail when it can't be located in a default directory.

A `--with-libunistring-prefix=` configure option would be appreciated. It also would add verbosity to this requirement.

Currently an ugly workaround via the CPPFLAGS/LDFLAGS/LIBS variables kludge is required to make things work:
```
./configure --prefix=/usr/local/targetdirectory/libpsl/0.21.2 \
    --enable-runtime=libidn2 \
    CPPFLAGS="-I/usr/local/targetdirectory/libunistring/latest/include" \
    LDFLAGS="-L/usr/local/targetdirectory/libunistring/latest/lib" \
    LIBS="-lunistring" \
    …
```

rockdaboot:
Fixes #214

@TurtleWilly @rb07 Can you test this please?

It looks like it works for me:
```
$ ./configure --enable-runtime=libidn2 --with-libunistring-prefix=/tmp
...
$ make
...
$ ldd tests/test-is-public
        linux-vdso.so.1 (0x00007ffd4179a000)
        libpsl.so.5 => /home/tim/src/libpsl/src/.libs/libpsl.so.5 (0x00007f2440e2c000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2440c2d000)
        libunistring.so.5 => /tmp/lib/libunistring.so.5 (0x00007f2440a70000)
        libidn2.so.0 => /lib/x86_64-linux-gnu/libidn2.so.0 (0x00007f2440a3d000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f2440e46000)
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -22,6 +22,9 @@ AC_C_INLINE
 dnl Check for visibility support
 gl_VISIBILITY
 
+dnl Unicodesupport
+gl_LIBUNISTRING
+
 #
 # Generate version defines for include file
 #
--- a/m4/absolute-header.m4
+++ b/m4/absolute-header.m4
@@ -1,0 +1,100 @@
+# absolute-header.m4 serial 18
+dnl Copyright (C) 2006-2024 Free Software Foundation, Inc.
+dnl This file is free software; the Free Software Foundation
+dnl gives unlimited permission to copy and/or distribute it,
+dnl with or without modifications, as long as this notice is preserved.
+
+dnl From Derek Price.
+
+# gl_ABSOLUTE_HEADER(HEADER1 HEADER2 ...)
+# ---------------------------------------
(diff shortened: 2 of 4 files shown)


rb07:
Yes, it works for me.

PR #241, opened 2024-03-24
Add prerequisites for building with GNU autotools

### LINKED ISSUE #157: maybe mention dependencies?
There are a couple of dependencies, for building from git, which are not mentioned in the README:

```
autoconf
autopoint
gtk-doc
```

rockdaboot:
Fixes #157

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,7 @@ Solaris OpenCSW [![Build Status Solaris amd64](https://buildfarm.opencsw.org/bui
 [![Build Status Solaris SparcV9](https://buildfarm.opencsw.org/buildbot/png?builder=libpsl-solaris10-sparcv9)](https://buildfarm.opencsw.org/buildbot/builders/libpsl-solaris10-sparcv9)
 
 
-libpsl - C library to handle the Public Suffix List
-===================================================
+# libpsl - C library to handle the Public Suffix List
 
 A *Public Suffix List* is a collection of Top Level Domains (TLDs) suffixes.
 TLDs include *Global Top Level Domains* (gTLDs) like `.com` and `.net`;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -124,24 +117,29 @@ Install with
...
 		cd libpsl
 
-Build with GNU autotools
+### Building with GNU autotools
+
+Prerequisites:
+  - python2.7+ (for converting the PSL list)
+  - basic C development tools (compiler, linker)
+  - git, make, autoconf, autoconf-archive, autopoint, automake, autotools,
+  - libtool, gettext, m4, pkg-config
```
eli-schwartz, line 133: A lot of this can be summarized as "need autotools". Mentioning autopoint and m4 in particular are very specific details that are really abstracted away by the primary projects they are bundled with or a dependency of -- m4 is required by autoconf since autoconf processes m4 files, autopoint is part of gettext.

README.md
```diff
@@ -124,24 +117,29 @@ Install with
...
+  - for building docs: gtk-doc-tools (gtkdocize)
+  - for building tarballs: gzip, lzip
```
eli-schwartz, line 135: building tarballs isn't part of building the project, it's part of maintaining the project. IMO it's not good material for README.md, but could be good material for RELEASING.md as project-internal release checklists etc.

README.md
rockdaboot: Indeed, `m4` has been pulled in as dependency here on Debian testing, but not so for `autopoint` nor `autotools-dev`. So I'd rather mention these tools/packages.

Removing `git` because without it you won't be able to fetch the sources!? (Mostly)

eli-schwartz: I did not realize autopoint was a separate package in Debian. How utterly pointless...

rockdaboot: It has been part of gettext a while ago, AFAIR. There must be some reason to separate both, but idea tbh.

PR #246, opened 2024-05-06
Add Libs.private to autoconf libpsl.pc

alyssais:
Static libraries don't include dependency information, so when using pkg-config to find out how to link with a static library, the .pc file for that library has to specify what other libraries to link. Libs.private is like Libs, except it's only used when static linking. This fixes linking against a static libpsl using pkg-config.

The Meson-generated libpsl.pc just includes all of its dependent libraries in Libs, so doesn't require any additional fix.


eli-schwartz (approved):
lgtm

I would personally reorganize it so the two Libs lines are together but that's just a nit.

PR #248, opened 2025-01-09
Support icu provided by Windows

paperchalice:
Windows provided `icu.lib` since [Windows 10 Version 1903](https://learn.microsoft.com/en-us/windows/win32/intl/international-components-for-unicode--icu-#version-1903-may-2019-update).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -15,6 +15,7 @@ notfound = dependency('', required : false)
 libidn2_dep = notfound
 libicu_dep = notfound
 libicucore_dep = notfound
+libicu_win_dep = notfound
 libidn_dep = notfound
 libunistring = notfound
 networking_deps = notfound
@@ -30,6 +31,17 @@ if enable_runtime == 'libicucore' or (host_machine.system() == 'darwin' and enab
     endif
   elif enable_runtime == 'libicucore'
     error('You requested libicucore but it is not available.')
+  endif
+endif
+
+if enable_runtime == 'libicu-win' or (host_machine.system() == 'windows' and enable_runtime == 'auto')
+  libicu_win_dep = cc.find_library('icu', has_headers : ['icu.h'], required: false)
+  if libicu_win_dep.found()
+    if enable_runtime == 'auto'
+      enable_runtime = 'libicu-win'
+    endif
+  elif enable_runtime == 'libicu-win'
+    error('You requested libicu-win but it is not available.')
   endif
 endif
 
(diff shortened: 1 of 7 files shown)


meson.build
```diff
@@ -40,10 +41,16 @@ if ['libidn2', 'auto'].contains(enable_runtime)
...
   libicu_dep = dependency('icu-uc', 'ICU',
                           components: 'uc',
                           required : false)
-  if libicu_dep.found()
+  if libicu_win_dep.found()
+    libicu_dep = libicu_win_dep
+    if enable_runtime == 'auto'
+      enable_runtime = 'libicu'
+    endif
+  elif libicu_dep.found()
```
paperchalice, line 53: Should there be an option here to force the use of external icu?

eli-schwartz: I'd say yes, since otherwise it sounds unpredictable to assume the resulting build will work. e.g you provide static ICU libraries in your build environment, upgrade your libpsl clone, and suddenly it starts linking to a different DLL shipped with only certain versions of Windows, so application distribution to older Windows systems breaks.

(Why can't Windows just ship actual forreal ICU...)

meson.build
```diff
@@ -57,6 +58,16 @@ if ['libicu', 'auto'].contains(enable_runtime)
   endif
 endif
 
+if ['libicu-win', 'auto'].contains(enable_runtime)
+  libicu_win_dep = cc.find_library('icu', required: false)
```
eli-schwartz, line 62: Please also skip this for "auto" when the host machine isn't Windows, as it can't be found anyway. Result: configuring on Linux or macOS or BSD with runtime == auto is a bit faster due to running one less check.

rockdaboot:
@paperchalice Sorry for the delay.
Was this done/working?
Can you rebase and see if you can extend the .github workflow for testing a windows build using icu?

paperchalice:
> @paperchalice Sorry for the delay. Was this done/working? Can you rebase and see if you can extend the .github workflow for testing a windows build using icu?

Windows workflow support is free, and it needs approval...

paperchalice:
> @paperchalice Sorry for the delay. Was this done/working? Can you rebase and see if you can extend the .github workflow for testing a windows build using icu?

Windows environment currently uses the mingw-w64 toolchain from strawberry perl, is this expected behavior?

eli-schwartz:
Most likely not. The Strawberry Perl install is a Frankensteined environment that isn't compatible with real mingw *or* with Microsoft libraries. Strawberry Perl absolutely refuses to stop installing their footgun to %PATH% even though it breaks many things, and GitHub installs Strawberry Perl so that you can run Perl scripts with perl.exe.

I'd say using `--vsenv` is the right move here indeed.

Note: meson can automatically search for a VS toolchain. The issue is that normally if you have gcc.exe or clang.exe on PATH, meson assumes you did that for a reason and skips activating VS. The `--vsenv` flag says "actually, regardless of what's in PATH I definitely want to activate VS".

eli-schwartz:
There are also dedicated actions which can set up vcvars for later steps, but given meson can handle it internally it may not be worth adding more external dependencies.

rockdaboot:
@eli-schwartz Is there anything blocking that prevents merging this PR? For me it lgtm, but I am neither a windows nor a meson expert.

Asking because I am about to draft a new release. The options for me are
- merge and possibly have a bugfix release soon
- not merge and have another release soon

Leaning towards the first option, as we will get faster feedback from users.

eli-schwartz:
I'm not a windows expert either, but the Meson changes look correct and we can see it detecting the newly implemented ICU detection in CI (and passing tests with it).

So it looks good to me.

rockdaboot:
TY for your great contribution @paperchalice!

PR #249, opened 2025-02-04
update psl list to latest version

lkounadis:


eli-schwartz:
I'm not sure what Tim's schedule for updating the PSL data is, but probably worth noting that it updates rather often and libpsl supports using your own copy of the file as a build option to `./configure` or `meson setup`. You can also configure it to load a separately distributed file possibly packaged e.g. as a Linux distro package to /user/share/psl-list.dat or whatever.

This is maybe a bit more convenient than regularly updating the submodule? YMMV...

lkounadis:
Thank you for the information! I'm currently working on getting this merged, hoping it will resolve a Safari bug caused by the outdated libpsl associated with this project. It is not clear to me if it will but it seems like so

Coeur (approved):

eli-schwartz:
Tim has just updated to https://github.com/publicsuffix/list/commit/64ba165cf391818a086139239c8fa223264eebcc

Issue #256, opened 2025-11-14
Statically link libpsl build with mingw-w64 13 break interop with previous mingw-w64 versions


JackCarterSmith:
This issue follow the one from libcurl (https://github.com/curl/curl/issues/19516).
I tried to build a test program with libcurl, which depend on libpsl, that I then tried to add statically.
That failed with an `undefined reference to 'stat64'`.
The libpsl.a have been build with MinGW-w64 **13** and the test program with MinGW-w64 **12**.

First and foremost, I'm not an expert with how libpsl use the `stat` element.
MinGW-w64 13 seem to break interop with previous MinGW versions which use `stat()` [function](https://github.com/X-EcutiOnner/curl-for-win/blob/ac25fe442126199a63bdd61b9de02bf0838da622/libpsl.sh#L52C7-L52C66).

Again, I don't know what should be the best way to handle/fix it.

JackCarterSmith:
Thanks for investigating. I forgot to precise I've used the pre-compiled libpsl.a from the windows 64-bits libcurl package.
I take a quick look on the log from your container, I suspect the line `CPPFLAGS="-I$INSTALLDIR/include"` can induce error in cross-building due to the possible presence of native headers files in the $INSTALLDIR/include location.
Hope that can help!

rockdaboot:
I did a cross-compilation with MinGW 13 on Debian testing, but can't reproduce the error yet
```
ii  mingw-w64-common                   13.0.0-1     all          Common files for Mingw-w64
ii  mingw-w64-x86-64-dev               13.0.0-1     all          Development files for MinGW-w64 targeting Win64
```

```
$ LDFLAGS="-L/usr/x86_64-w64-mingw32/lib" ./configure --host=x86_64-w64-mingw32 --disable-runtime RUNTEST='wine'
$ make
$ make check
...
make[4]: Entering directory '/home/tim/src/libpsl/tests'
PASS: test-is-public-builtin.exe
PASS: test-registrable-domain.exe
PASS: test-is-cookie-domain-acceptable.exe
PASS: test-is-public.exe
PASS: test-is-public-all.exe
...
```

JackCarterSmith:
From what I've understand, is when the libpsl is build with MinGW 13 and the test program is build with MinGW 12 (with static libpsl.a) that the linking step of the test program crash (with undefined 'stat64' error).
I suppose the tests programs from Debian testing container have been built against MinGW 13.

[...]

PR #259, opened 2025-11-18
Updated stat sys function call support on Windows

JackCarterSmith:
Following https://github.com/rockdaboot/libpsl/issues/256.

## Why?

I propose a quick patch to directly use crt function _stat() on Windows platform instead of the interface provided by MinGW to avoid breaking compatibility with change in MinGW versions.
As stat() function signature didn't change, it's only a rework of the abstract interface of MinGW who cause a compatibility issue (and not a change in the CRT).

## Regression testing

I've tested the changes in this PR using `make check` on a Ubuntu 22.04 platform with GCC-14 for native linux and with GCC-14 MinGW-w64 13 in cross-compilation.
Plus I tested on a test program directly on Windows using GCC-14 MinGW-w64 **12** natively.

## What does this patch involve?

This patch only change the stat() and struct stat redirection from MinGW to direct CRT (similar to how libcurl does it) only on Windows build. The linux sys/stat() is keep untouched.
The stat() symbol in the otutput .a lib on linux is still `stat`.
On Windows, the previous glitch generated a `stat64` symbol, which [aren't pushed in the tagged v13](https://github.com/mingw-w64/mingw-w64/issues/135#issuecomment-3541489769) of MinGW.
Now, with this patch, the symbol is `_imp__stat64` instead and use the Windows UCRT .dll directly.

### DIFF
--- a/src/psl.c
+++ b/src/psl.c
@@ -38,6 +38,16 @@
 #       define GCC_VERSION_AT_LEAST(major, minor) 0
 #endif
 
+/* Must be defined before <sys/stat.h> */
+#if defined(_MSC_VER) || defined(__MINGW32__)
+# define USE_WIN32_LARGE_FILES
+# ifdef __MINGW32__
+#   ifndef _FILE_OFFSET_BITS
+#    define _FILE_OFFSET_BITS 64
+#   endif
+# endif
+#endif
+
 #include <sys/types.h>
 #include <sys/stat.h>
 
@@ -46,6 +56,21 @@
 # define WIN32_LEAN_AND_MEAN
 # endif
 # include <windows.h> /* for GetACP() */
+#endif
+
+#if defined(_WIN32)
+# ifdef USE_WIN32_LARGE_FILES
+#  define struct_stat  struct _stati64
+#  define func_sys_stat _stati64
+# else
+#  define struct_stat  struct _stat
+#  define func_sys_stat _stat
+# endif
+#endif
+
+#ifndef struct_stat
+# define struct_stat   struct stat
+# define func_sys_stat stat
 #endif
 
 #if defined(_MSC_VER) && ! defined(ssize_t)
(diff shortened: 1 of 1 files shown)


JackCarterSmith:
Since it's been a while, I "up" this PR.

JackCarterSmith:
Sorry to bother you @eli-schwartz, maybe I should have ping-ed the relevant person :)

eli-schwartz:
Sorry, it's been a bit busy here recently so a direct ping is definitely a more reliable way to attract my attention. :D

I'm not a windows expert but this seems like a plausible way to use the underscore variant on mingw (which is what you want).

I might reword the commit message, mainly because I'm a kind of nitpicky person about commit messages.


> avoid mingw stat wrapper on Windows
> 
> stat() is a mingw posix wrapper for the native Windows _stat() and friends. Although tempting to use for simplicity, the wrapper incurs overhead, such as making distribution of static libraries challenging (as the mingw version has to match exactly). Additionally some mingw releases are buggy and generate symbol linkage to missing symbols. Avoid the middleman and utilize the windows API directly.

eli-schwartz:
And a general commentary on "update XYZ support" -- a commit message one-line summary should ideally be optimized to tell the reader as concisely as possible, what the intended outcome is. "Update XYZ" is not the goal, it's incidental to the goal of "avoid a potentially unreliable wrapper". Another possible description might be "make mingw static linking more robust".

JackCarterSmith:
I don't mean to rush you! I sometimes forget things and lose track of time either :D

No problem with that—I completely agree with the principle and the content. I force-pushed the reword just as it is.

Thank you for the review and the commit tip!

lhmouse:
mingw-w64 provides `stat()` with some overhead to work around bugs in `_stat()` in old MSVCR* DLLs, for example the latter may flag directories and device files as regular, and do not accept UNC paths which start with `\\?\`.

If you always link against UCRT you may ignore differences between `stat()` and `_stat()`, since both reference the same function. (and `stat64()` and `_stat64()`, and others likewise.)

JackCarterSmith:
@lhmouse Thanks for the clarification!

If I understand correctly, the only stat() calls are used to retrieve the timestamp of the “fixed” psl file.
I suppose the risk of compatibility issues between versions, as you mentioned, is limited, is that right?

lhmouse:
> If I understand correctly, the only stat() calls are used to retrieve the timestamp of the “fixed” psl file.
> 
> I suppose the risk of compatibility issues between versions, as you mentioned, is limited, is that right?

Yes, I think so.

Issue #262, opened 2026-05-05
Improve MSVC support with Autotools


maiddaisuki:
Hello. A few months ago I reached out to Tim with a few suggestion on how to improve MSVC support when using Autotools. I've prepared a series of changes which address those issues. Here's the summary of issues I've encountered; note that not all of them are MSVC-specific.

### `make check` does not work with out-of-tree build from git checkout

When I tried to run `make check` after building libpsl, I saw the following error:

```
/usr/share/automake-1.18/compile clang-cl.exe -DHAVE_CONFIG_H -I. -I/h/git/libpsl/fuzz -I..  -I/h/git/libpsl/include -I/h/git/libpsl/fuzz -DSRCDIR=\"/h/git/libpsl/fuzz\" -DTEST_RUN ... - -c -o libpsl_fuzzer.obj `cygpath -w '/h/git/libpsl/fuzz/libpsl_fuzzer.c'`
H:\git\libpsl\fuzz\libpsl_fuzzer.c(42,10): fatal error: 'libpsl.h' file not found
   42 | #include "libpsl.h"
      |          ^~~~~~~~~~
1 error generated.
```

I found the issue to be in [fuzz/Makefile.am](https://github.com/rockdaboot/libpsl/blob/8f769dd06d3ac888dcd0ec2a8cfc46e3f00a413c/fuzz/Makefile.am):

```
AM_CPPFLAGS = -I$(top_srcdir)/include -I$(srcdir) -DSRCDIR=\"$(abs_srcdir)\"
```

and in [tests/Makefile.am](https://github.com/rockdaboot/libpsl/blob/8f769dd06d3ac888dcd0ec2a8cfc46e3f00a413c/tests/Makefile.am):

```
AM_CPPFLAGS = -I$(top_srcdir)/include
```

Theirs `AM_CPPFLAGS` contain -I flags only for the source tree, while libpsl.h is generated in the build tree.

After fixing it locally, I run `make check` again and I see all tests pass except for `test-is-public-all`. The test-suite.log contains the following:

```
FAIL: test-is-public-all
========================

Failed to load 'H:/git/libpsl/tests/psl.dafsa'
Failed to load 'H:/git/libpsl/tests/psl_ascii.dafsa'
loaded 10041 suffixes and 8 exceptions
builtin PSL has 10041 suffixes and 8 exceptions
Summary: 2 out of 189884 tests failed
```

[...]

PR #263, opened 2026-05-05
Improve MSVC support with Autotools

maiddaisuki:
See #262.

I'm not sure about correctness of "configure.ac: check for Solaris networking libraries early"; I don't have access to Solaris to test it. 

Does libpsl really need those networking libraries? I saw that in meson.build, we add -lws2_32 on Windows, but configure does not do it and Windows builds still work.

I tested these changes with MSVC using my own [build script](https://github.com/maiddaisuki/gnu-libs-with-msvc), including configuration when using static CRT (/MT) and only static dependencies.

I also tested these changes with mingw-w64 (using Msys2) and on my Debian machine. In both cases I tried all three runtimes (libidn2, libidn and libicu) and all configurations worked properly.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -213,6 +213,12 @@ AC_SUBST([AM_CPPFLAGS])
 AC_SUBST([AM_CFLAGS])
 AC_SUBST([AM_LDFLAGS])
 
+dnl If we are building only static libraries, we need to define PSL_STATIC
+dnl when building psl executable; this is required on Windows since libpsl
+dnl uses __declspec(dllimport) by default.
+AS_IF([test "x${enable_shared}" = "xno"],
+  [AS_VAR_APPEND([AM_CPPFLAGS], [" -DPSL_STATIC"])])
+
 dnl We use pkg-config/pkgconf to find dependencies. We want to ensure that we
 dnl are able to find them even if they are installed as static libraries only.
 dnl
--- a/configure.ac
+++ b/configure.ac
@@ -297,9 +297,17 @@ if test "$enable_runtime" = "auto"; then
   enable_runtime=no
 fi
 
-dnl If we use libidn or libidn2, we also need libunistring
+dnl If we use libidn or libidn2, we also need libiconv and libunistring
 if test "x$enable_runtime" = "xlibidn2" -o "x$enable_runtime" = "xlibidn"; then
   __save_LIBS=$LIBS
+
+  LIBS=$LIBICONV
+  AC_LINK_IFELSE([
+    AC_LANG_PROGRAM([#include <iconv.h>], [iconv_open ("", "");])
+  ], [:], [
+    AC_MSG_ERROR([You requested libidn2|libidn but libiconv is not installed.])
+  ])
+  LIBS=$__save_LIBS
 
   LIBS=$LIBUNISTRING
   AC_LINK_IFELSE([
--- a/configure.ac
+++ b/configure.ac
@@ -190,6 +190,19 @@ if test "$enable_builtin" = "yes"; then
   AC_DEFINE([ENABLE_BUILTIN], [1], [Generate built-in PSL data])
 fi
 
+dnl Global project-specific compiler and linker flags.
+dnl
+dnl We should not use plain *FLAGS since they belong to the user.
+dnl We can temporary modify them for configuration checks, but then we need to
+dnl restore them to their original values.
+AM_CPPFLAGS="-I\$(top_builddir)/include -I\$(top_srcdir)/include"
+AM_CFLAGS=
+AM_LDFLAGS=
+
+AC_SUBST([AM_CPPFLAGS])
+AC_SUBST([AM_CFLAGS])
+AC_SUBST([AM_LDFLAGS])
+
 if test "$enable_runtime" = "libidn2" -o "$enable_runtime" = "auto"; then
   # Check for libidn2
   PKG_CHECK_MODULES([LIBIDN2], [libidn2], [
--- a/configure.ac
+++ b/configure.ac
@@ -22,6 +22,12 @@ AC_C_INLINE
 dnl Check for visibility support
 gl_VISIBILITY
 
+dnl Solaris has socket in libsocket and inet_ntop in libnsl, but also needs
+dnl libsocket, so the order is important here
+AC_CHECK_LIB([socket], [socket], [
+  AC_CHECK_LIB([nsl], [inet_ntop], [:], [:])
+], [:])
+
 dnl AM_ICONV sets @LIBICONV@ and @LTLIBICONV@ for use in Makefile.am
 dnl libunistring may depend on libiconv, so invoke it before gl_LIBUNISTRING
 AM_ICONV
@@ -322,19 +328,6 @@ AM_CONDITIONAL([WITH_LIBICU], test "x$enable_runtime" = "xlibicu")
 AM_CONDITIONAL([WITH_LIBIDN2], test "x$enable_runtime" = "xlibidn2")
 AM_CONDITIONAL([WITH_LIBIDN], test "x$enable_runtime" = "xlibidn")
 AM_CONDITIONAL([ENABLE_BUILTIN], test "x$enable_builtin" = "xyes")
-
-# Solaris has socket in libsocket and inet_ntop in libnsl, but also needs libsocket, so the order is important here
-AC_CHECK_LIB([socket], [socket], [NEEDS_SOCKET=yes], [])
-if test -n "$NEEDS_SOCKET" ; then
-  AC_CHECK_LIB([nsl], [inet_ntop], [NEEDS_NSL=yes], [])
-fi
-if test -n "$NEEDS_SOCKET" && test -n "$NEEDS_NSL" ; then
-  LIBS="$LIBS -lsocket -lnsl"
-elif test -n "$NEEDS_SOCKET" ; then
-  LIBS="$LIBS -lsocket"
-elif test -n "$NEEDS_NSL" ; then
-  LIBS="$LIBS -lnsl"
-fi
 
 # Check for clock_gettime() used for performance measurement
 AC_SEARCH_LIBS(clock_gettime, rt)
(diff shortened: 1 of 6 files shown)


rockdaboot:
Great work @maiddaisuki! Overall LGTM.

I like that you split the PR into several commits :+1:
This allows me to review and merge them one after the other (non-debatable things first).

Please see if you can fix the libicu test in CI (~works for me locally on Debian testing~).

I can reproduce the issue with
```
./configure --enable-runtime=libicu
make clean && make
```

This is a work-around, but I think something in `libpsl.la` is missing:
```diff
diff --git a/fuzz/Makefile.am b/fuzz/Makefile.am
index 8d31f90..b2987ae 100644
--- a/fuzz/Makefile.am
+++ b/fuzz/Makefile.am
@@ -13,6 +13,7 @@ if WITH_LIBICU
   libpsl_icu_fuzzer_SOURCES = libpsl_fuzzer.c $(MAIN)
   libpsl_icu_load_fuzzer_SOURCES = libpsl_load_fuzzer.c $(MAIN)
   libpsl_icu_load_dafsa_fuzzer_SOURCES = libpsl_load_dafsa_fuzzer.c $(MAIN)
+  libpsl_icu_fuzzer_LDADD = $(LDADD) $(LIBICU_LIBS)
   PSL_TESTS = \
    libpsl_icu_fuzzer$(EXEEXT) \
    libpsl_icu_load_fuzzer$(EXEEXT) \
```

Maybe better
```diff
diff --git a/fuzz/Makefile.am b/fuzz/Makefile.am
index 8d31f90..caefe0e 100644
--- a/fuzz/Makefile.am
+++ b/fuzz/Makefile.am
@@ -8,18 +8,19 @@ if !FUZZING
 endif
 
 if WITH_LIBICU
-  XLIBS = $(LIBICU_LIBS)
+  XLIBS = $(runtime_LIBS)
   XTYPE = _icu
   libpsl_icu_fuzzer_SOURCES = libpsl_fuzzer.c $(MAIN)
   libpsl_icu_load_fuzzer_SOURCES = libpsl_load_fuzzer.c $(MAIN)
   libpsl_icu_load_dafsa_fuzzer_SOURCES = libpsl_load_dafsa_fuzzer.c $(MAIN)
+  libpsl_icu_fuzzer_LDADD = $(LDADD) $(runtime_LIBS)
   PSL_TESTS = \
    libpsl_icu_fuzzer$(EXEEXT) \
    libpsl_icu_load_fuzzer$(EXEEXT) \
    libpsl_icu_load_dafsa_fuzzer$(EXEEXT)
 else
 if WITH_LIBIDN2
-  XLIBS = -lidn2 -lunistring
+  XLIBS = $(runtime_LIBS)
   XTYPE = _idn2
   libpsl_idn2_fuzzer_SOURCES = libpsl_fuzzer.c $(MAIN)
   libpsl_idn2_load_fuzzer_SOURCES = libpsl_load_fuzzer.c $(MAIN)
@@ -30,7 +31,7 @@ if WITH_LIBIDN2
    libpsl_idn2_load_dafsa_fuzzer$(EXEEXT)
 else
 if WITH_LIBIDN
-  XLIBS = -lidn -lunistring
+  XLIBS = $(runtime_LIBS)
   XTYPE = _idn
   libpsl_idn_fuzzer_SOURCES = libpsl_fuzzer.c $(MAIN)
   libpsl_idn_load_fuzzer_SOURCES = libpsl_load_fuzzer.c $(MAIN)
```

configure.ac
```diff
@@ -22,6 +22,12 @@ AC_C_INLINE
 dnl Check for visibility support
 gl_VISIBILITY
 
+dnl Solaris has socket in libsocket and inet_ntop in libnsl, but also needs
+dnl libsocket, so the order is important here
+AC_CHECK_LIB([socket], [socket], [
```
rockdaboot, line 27: I can no longer test on Solaris, can/did you?
But even if not and if these changes break anything, someone will show up who can test.

configure.ac
```diff
@@ -190,88 +200,287 @@ if test "$enable_builtin" = "yes"; then
...
   if test "x$HAVE_LIBIDN2" = "xyes"; then
+    runtime_CFLAGS=$LIBIDN2_CFLAGS
+    runtime_LIBS=$LIBIDN2_LIBS
+
     if test "$enable_runtime" = "auto"; then
       enable_runtime=libidn2
       AC_DEFINE([WITH_LIBIDN2], [1], [generate PSL data using libidn2])
     fi
+  else
+    if test "$enable_runtime" = "libidn2"; then
```
rockdaboot, line 309: Reminds me of fixing `if test` into `AS_IF`

maiddaisuki:
> Please see if you can fix the libicu test in CI (~works for me locally on Debian testing~).

It would help to configure with `--disable-silent-rules` during CI runs to see the command line.

> I can reproduce the issue with
> 
> ```
> ./configure --enable-runtime=libicu
> make clean && make
> ```
> 
> This is a work-around, but I think something in `libpsl.la` is missing:

Hm, I expected libtool to pull in linker flags from libpsl.la, but it seems like it doesn't? It worked for me on Debian and Msys2, but it is possible that older libtool versions are buggy; I was building from archive created with `make dist`, so I was using latest libtool.

> ```diff
> diff --git a/fuzz/Makefile.am b/fuzz/Makefile.am
> index 8d31f90..b2987ae 100644
> --- a/fuzz/Makefile.am
> +++ b/fuzz/Makefile.am
> @@ -13,6 +13,7 @@ if WITH_LIBICU
>    libpsl_icu_fuzzer_SOURCES = libpsl_fuzzer.c $(MAIN)
>    libpsl_icu_load_fuzzer_SOURCES = libpsl_load_fuzzer.c $(MAIN)
>    libpsl_icu_load_dafsa_fuzzer_SOURCES = libpsl_load_dafsa_fuzzer.c $(MAIN)
> +  libpsl_icu_fuzzer_LDADD = $(LDADD) $(LIBICU_LIBS)
>    PSL_TESTS = \
>     libpsl_icu_fuzzer$(EXEEXT) \
>     libpsl_icu_load_fuzzer$(EXEEXT) \
> ```
> 
> Maybe better
> 
> ```diff
> diff --git a/fuzz/Makefile.am b/fuzz/Makefile.am
> index 8d31f90..caefe0e 100644
> --- a/fuzz/Makefile.am
> +++ b/fuzz/Makefile.am
> @@ -8,18 +8,19 @@ if !FUZZING
>  endif
>  
>  if WITH_LIBICU
> -  XLIBS = $(LIBICU_LIBS)
> +  XLIBS = $(runtime_LIBS)
>    XTYPE = _icu
>    libpsl_icu_fuzzer_SOURCES = libpsl_fuzzer.c $(MAIN)
>    libpsl_icu_load_fuzzer_SOURCES = libpsl_load_fuzzer.c $(MAIN)
>    libpsl_icu_load_dafsa_fuzzer_SOURCES = libpsl_load_dafsa_fuzzer.c $(MAIN)
> +  libpsl_icu_fuzzer_LDADD = $(LDADD) $(runtime_LIBS)
>    PSL_TESTS = \
>     libpsl_icu_fuzzer$(EXEEXT) \
>     libpsl_icu_load_fuzzer$(EXEEXT) \
>     libpsl_icu_load_dafsa_fuzzer$(EXEEXT)
>  else
>  if WITH_LIBIDN2
> -  XLIBS = -lidn2 -lunistring
> +  XLIBS = $(runtime_LIBS)
>    XTYPE = _idn2
>    libpsl_idn2_fuzzer_SOURCES = libpsl_fuzzer.c $(MAIN)
>    libpsl_idn2_load_fuzzer_SOURCES = libpsl_load_fuzzer.c $(MAIN)
> @@ -30,7 +31,7 @@ if WITH_LIBIDN2
>     libpsl_idn2_load_dafsa_fuzzer$(EXEEXT)
>  else
>  if WITH_LIBIDN
> -  XLIBS = -lidn -lunistring
> +  XLIBS = $(runtime_LIBS)
>    XTYPE = _idn
>    libpsl_idn_fuzzer_SOURCES = libpsl_fuzzer.c $(MAIN)
>    libpsl_idn_load_fuzzer_SOURCES = libpsl_load_fuzzer.c $(MAIN)
> ```

I was considering replacing hardcoded -l flags with `$(runtime_CFLAGS)`, but I was afraid to break `oss-fuzz` rule which has

```
	-Wl,-Bstatic \
	$(XLIBS) \
	-Wl,-Bdynamic; \
```

configure.ac
maiddaisuki: I do not have access to Solaris as well, so no, I did not test it.

configure.ac
maiddaisuki: I also noticed that some macro invocations in configure.ac are missing proper quoting. I was thinking to send a PR for this and some other clean ups in configure.ac once we're done with this PR.

maiddaisuki:
> Please see if you can fix the libicu test in CI (~works for me locally on Debian testing~).
> 
> I can reproduce the issue with
> 
> ```
> ./configure --enable-runtime=libicu
> make clean && make
> ```

So, on Debian

```
$ libtoolize --version
```

reports libtool 2.4.7 which is pretty old, and by building with `make V=1` I can see that it is really does not pull dependencies from libpsl.la; it does so with latest libtool, so it seems to be old buggy libtool to blame.

I think the simplest fix is to add `$(runtime_LIBS)` to `LDADD` in fuzz/Makefile.am. I tested it locally and it fixes `make check`. What do you think?

rockdaboot:
> > Please see if you can fix the libicu test in CI (~works for me locally on Debian testing~).
> > I can reproduce the issue with
> > ```
> > ./configure --enable-runtime=libicu
> > make clean && make
> > ```
> 
> So, on Debian
> 
> ```
> $ libtoolize --version
> ```
> 
> reports libtool 2.4.7 which is pretty old, and by building with `make V=1` I can see that it is really does not pull dependencies from libpsl.la; it does so with latest libtool, so it seems to be old buggy libtool to blame.

:+1: Nice finding!

> I think the simplest fix is to add `$(runtime_LIBS)` to `LDADD` in fuzz/Makefile.am. I tested it locally and it fixes `make check`. What do you think?

Sounds good to me!

rockdaboot:
Thank you so much, @maiddaisuki :+1:

configure.ac
rockdaboot: That would be highly appreciated!

PR #270, opened 2026-06-23
Support libicucore on Apple OSes

Bo98:
Apple OSes (macOS etc) ship with a libicucore with the system. For the purposes of libpsl, this is basically the same as regular ICU except that it doesn't have `ucnv` and requires a different `-l` flag.

This PR adds support for libicucore by using the ICU paths, except for `psl_str_to_utf8lower` where it uses the iconv path but with ICU calls instead of libunistring.

The benefit here is that on macOS, `libpsl` now no longer requires any external dependencies beyond what ships with the OS.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -153,6 +153,7 @@ AC_ARG_ENABLE(runtime,
       Specify the IDNA library used for libpsl run-time conversions:
         libidn2 [[default]]: IDNA2008 library (also needs libunistring)
         libicu:            IDNA2008 UTS#46 library
+        libicucore:        IDNA2008 UTS#46 library shipped with Apple OSes
         libidn:            IDNA2003 library (also needs libunistring)
   --disable-runtime        Do not link runtime IDNA functionality
   ], [
@@ -162,6 +163,9 @@ AC_ARG_ENABLE(runtime,
     elif test "$enableval" = "libicu"; then
       enable_runtime=libicu
       AC_DEFINE([WITH_LIBICU], [1], [generate PSL data using libicu])
+    elif test "$enableval" = "libicucore"; then
+      enable_runtime=libicucore
+      AC_DEFINE([WITH_LIBICUCORE], [1], [generate PSL data using libicucore])
     elif test "$enableval" = "libidn"; then
       enable_runtime=libidn
       AC_DEFINE([WITH_LIBIDN], [1], [generate PSL data using libidn])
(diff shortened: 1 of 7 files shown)


configure.ac
```diff
@@ -489,27 +518,29 @@ if test "$enable_runtime" = "auto"; then
...
+      AC_LANG_PROGRAM([char u8_tolower (void);], [u8_tolower ();])
+    ], [:], [
+      AC_MSG_ERROR([You requested libidn2|libidn but libunistring is not installed.])
+    ])
+    LIBS=$__save_LIBS
+    runtime_LIBS="$runtime_LIBS $LTLIBUNISTRING"
+  fi
 fi
 
 AM_CONDITIONAL([WITH_LIBICU], test "x$enable_runtime" = "xlibicu")
```
rockdaboot, line 546: We should have a `AM_CONDITIONAL([WITH_LIBICUCORE] ...` here as well.

.gitignore
```diff
@@ -63,13 +63,15 @@ Makefile
 Makefile.in
 missing
 nbproject/
+po/fetch-po
```
rockdaboot: nit: drop the unrelated changes to `.gitignore` in favor of a separate PR

Bo98:
Pushed a fix for that `libidn` CI error - was a small typo

rockdaboot (commented):
TY for your contribution! Just some minor comments.

Would you like to mention libicucore in the README.md?
Can you add libicucore to the CI test matrix (atm: `        runtime: ["libidn", "libidn2", "libicu"]`?

.gitignore
Bo98: Ah forgot about these - will do!

configure.ac
Bo98: Done

Bo98:
> Can you add libicucore to the CI test matrix (atm: ` runtime: ["libidn", "libidn2", "libicu"]`?

I missed this bit, sorry. Looks like meson macOS test already covers it. I can add it to the autoconf path too, though will also need to use the macOS runner for that runtime.

rockdaboot:
> > Can you add libicucore to the CI test matrix (atm: ` runtime: ["libidn", "libidn2", "libicu"]`?
> 
> I missed this bit, sorry. Looks like meson macOS test already covers it. I can add it to the autoconf path too, though will also need to use the macOS runner for that runtime.

No stress, when it's covered by meson. I'll leave it to you to extend the matrix or not, can be done in a follow-up as well.

I'll wait with merging until tomorrow morning, so there is time adding libicucore to `README.md` (trivial change).

Bo98:
All set up and working now. Also adjusted the README, feel free to tweak it further if you'd like.

Thanks for the reviews!

rockdaboot (approved):
TY for the great work!

PR #271, opened 2026-06-23
Fix handling of dafsa tests with in-tree builds

Bo98:
`$srcdir` can be a relative directory which is not safe to encode into the makefile given it runs tests under a subdirectory. Instead encode `$(top_srcdir)` in a similar fashion we do to `$(top_builddir)`.


rockdaboot:
Please pardon me, I am in favor of #272, which drops the path definitions from configure.ac entirely. But I am happy to get your opinion on that.

Bo98:
I agree dropping it from configure.ac is better. Thanks!

PR #273, opened 2026-06-23
Update .gitignore for newer gettext

Bo98:
`insert-header.sed` was added in gettext 0.23 and fetch-po was added in gettext 1.0.

We could probably do this in a better way (e.g. invert it so it ignores everything but the two files checked in), but this PR is the minimum required change for building with newer gettext.


rockdaboot:
TY for your contribution!

PR #281, opened 2026-06-25
Properly handle trailing dot in domains

### LINKED ISSUE #279: psl_is_cookie_domain_acceptable does not accept a trailing dot
So checking `co.uk.` for example says it *is* an acceptable cookie domain.

This is not a documented restriction, so I suggest either the function is modified to ignore a single trailing dot or it gets clarified in the documentation that this kind of strip needs to be done by the caller. I personally prefer the former: I think it should ignore them.

curl's vulnerability [CVE-2026-8924](https://curl.se/docs/CVE-2026-8924.html) happened which made me realize this

rockdaboot:
Accept a single trailing dot (e.g., `co.uk` and `co.uk.` are treated equally -> public suffix).

Fixes #279

### DIFF
--- a/src/psl.c
+++ b/src/psl.c
@@ -312,7 +312,7 @@ static int suffix_compare(const psl_entry_t *s1, const psl_entry_t *s2)
 	if ((n = s1->length - s2->length))
 		return n;  /* shorter rules first */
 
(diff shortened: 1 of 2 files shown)


bagder (approved):

bagder:
> Accept a single leading dot (e.g., co.uk and co.uk. are treated equally -> public suffix).

*trailing* though, not leading

PR #285, opened 2026-07-05
Simplify license headers

### LINKED ISSUE #284: Copyright year inconsistency: main 2014-2022, src files end at 2015
The Debian copyright file for libpsl shows:
- Main (Files: *): Copyright 2014-2022
- Files: src/lookup_string_in_fixed_set.c: Copyright 2014-2015
- Files: src/psl-make-dafsa: Copyright 2014-2015

Some source files have copyright ending at 2015 while the main project copyright goes to 2022. Please review and update if needed.

rockdaboot:
## Problem
Maintenance burden due to having copyright years in every source code file.

## Suggestion to reduce maintenance burden
- remove copyright years and instead refer to git history
- remove license text from files and refer to LICENSE file
- add SPDX-License-Identifier to files

## Remark
This PR is not complete yet.
If/When we find consent, I'll amend all files appropriately.
Please let me know what you think, @eli-schwartz

Fixes #284


rockdaboot:
The git reference doesn't apply to tarballs. Maybe we need the copyright years in LICENSE and COPYING.

eli-schwartz:
It is possible to have gitattributes `export-subst` replace a template with information from a git commit. This is often used to do stuff like set version numbers in `git-archive` tarballs based on the git tag associated with the archive.

It works with git autogenerated "source code" tarballs, which can cover people downloading the wrong release asset. And you can also have `make distcheck` write out this information.

So, if you want to avoid manually updating the copyright year, that's what I'd do.

rockdaboot:
> It is possible to have gitattributes `export-subst` replace a template with information from a git commit. This is often used to do stuff like set version numbers in `git-archive` tarballs based on the git tag associated with the archive.
> 
> It works with git autogenerated "source code" tarballs, which can cover people downloading the wrong release asset. And you can also have `make distcheck` write out this information.
> 
> So, if you want to avoid manually updating the copyright year, that's what I'd do.

Thanks @eli-schwartz, just learned about `export-subst`.

There are tools like license checkers that read `LICENSE` etc. and other files from git. Not sure how they deal with a format/template instead of a concrete date/year.

Atm, I'd prefer not using templates and manually update the copyright year at max once a year just in the `LICENSE` file. Since we already use `AM_INIT_AUTOMAKE([foreign])`, we can safely remove `COPYING` (it is identical to `LICENSE`). And follow the points mentioned in the issue description.

Not set in stone though, we can improve in the future using templates, if we feel confident about the possible corner cases.

PR #286, opened 2026-07-06
Always add default test files to the tarball

chenx97:
Currently, `make dist` only adds the test file specified by `PSL_TESTFILE`. This patch unconditionally adds both tests from list/tests, while `PSL_TESTFILE` can still be set to add or overwrite a desired test file.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -26,6 +26,7 @@ EXTRA_DIST = build-aux/config.rpath LICENSE meson.build  meson_options.txt
 dist-hook:
 	mkdir -p $(distdir)/list/tests
 	cp -p $(PSL_FILE) $(distdir)/list
+	cp -p $(PSL_DEFAULT_TESTFILES) $(distdir)/list/tests
 	cp -p $(PSL_TESTFILE) $(distdir)/list/tests
 
 clean-local:
--- a/configure.ac
+++ b/configure.ac
@@ -619,6 +619,8 @@ AC_ARG_WITH([psl-testfile],
   [PSL_TESTFILE=$withval],
   [PSL_TESTFILE="\$(top_srcdir)/list/tests/tests.txt"])
 AC_SUBST([PSL_TESTFILE])
+AC_SUBST([PSL_DEFAULT_TESTFILES],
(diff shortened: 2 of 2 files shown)


rockdaboot:
Can you explain why `list/tests/test_psl.txt` is needed in the tarball?

chenx97:
> Can you explain why `list/tests/test_psl.txt` is needed in the tarball?

It allows tarball users to verify their build against both formats of the same test.

rockdaboot:
> > Can you explain why `list/tests/test_psl.txt` is needed in the tarball?
> 
> It allows tarball users to verify their build against both formats of the same test.

But why would anyone want this? It's semantically the same content. But the libpsl tests do not support this format. So it's of no use at the moment.

chenx97:
> But the libpsl tests do not support this format.

https://github.com/rockdaboot/libpsl/blob/master/tests%2Ftest-registrable-domain.c#L161-L166

It seems to have limited support, but at least it works in this test.

rockdaboot:
> > But the libpsl tests do not support this format.
> 
> https://github.com/rockdaboot/libpsl/blob/master/tests%2Ftest-registrable-domain.c#L161-L166
> 
> It seems to have limited support, but at least it works in this test.

Ok right, it does as a fallback case. So someone could provide his own/custom tests.

But you suggest to add a redundant file. Why would anyone want this? No value added, it's just noise / bloat. Do you have a concrete use case?

chenx97:
> Do you have a concrete use case?

This PR's inspired by an attempt to transition the packaging of libpsl from the git source to the tarball for a Linux distribution, where test_psl.txt exists in the current package.

We discussed your question and realized that the file was there for no other reason than that ["Debian has it, so we'd better have it as well"](https://github.com/AOSC-Dev/aosc-os-abbs/blob/c7846757a765bf3f122f40501bc6721fbad9edba/runtime-network/libpsl/autobuild/beyond#L13). I admit that this isn't really a concrete use case because we can safely remove this file from our distribution without breaking anything.

rockdaboot:
Thanks for letting me know!

PR #288, opened 2026-07-13
meson: also rename cdata to _cdata in docs/libpsl/meson.build

florianernst:
Otherwise, when building the docs using meson, the build will break with `../docs/libpsl/meson.build:7:20: ERROR: Unknown variable "cdata".` now that `cdata`  has been renamed to `_cdata` in da5479661839e25783ee482e0180efc3d28f2428.


rockdaboot (approved):

rockdaboot:
Thanks for your contribution!

PR #292, opened 2026-07-25
Allow explicitly disabling fuzzing

thomasbruggink:
Using `--disable-fuzzing` explicitly to create fixed flag builds currently turns on fuzzing since configure.ac manually overrides the flag.
Since the enable-fuzzing flag is set by configure it should confirm its state is true. This is already done for other flags but not for the fuzzing flag.

This PR corrects this so `--disable-fuzzing` works as expected.

Before:
```
./configure --disable-fuzzing
cat config.log | grep -i "fuzzing build"
> Fuzzing build:     yes,
```
After:
```
./configure --disable-fuzzing
cat config.log | grep -i "fuzzing build"
> Fuzzing build:     no,
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -94,9 +94,10 @@ PKG_PROG_PKG_CONFIG
 
 AC_ARG_ENABLE([fuzzing],
   [AS_HELP_STRING([--enable-fuzzing], [Turn on fuzzing build (for developers)])], [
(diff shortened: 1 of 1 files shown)


rockdaboot:
Thanks, nice finding!

rockdaboot (approved):

PR #293, opened 2026-07-25
Use python3 in the psl-make-dafsa shebang

looptc:
The script is run through `#!/usr/bin/env python`, which fails on systems that
no longer ship a `python` binary. Since Python 2 reached end of life most
distributions provide only `python3`, and PEP 394 recommends that scripts
requiring Python 3 spell it out.

The script itself is already Python 3 (it is invoked as `python3` by
src/meson.build), so this only fixes running it directly.
Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/src/psl-make-dafsa
+++ b/src/psl-make-dafsa
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # Copyright 2014 The Chromium Authors. All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE.chromium file.


rockdaboot:
TY for your suggestion.

My thoughts are, PEP or not, this is a breaking change. The project's code is legacy (C89), and as long there is no real-world problem with it, let's keep `python`. Debian for example has a light-weight package `python-is-python3` that solves the issue for users.

When time comes to modernize the code, I'll merge his PR. So let's keep it open.

looptc:
Understood, and thanks for keeping it open.

One data point for the "no real-world problem with it" part: we build a from-source system where there is no `python` binary at all — not a stance about PEP 394, just minimalism (only `python3` gets installed). `python-is-python3` is a Debian package, so on a source-based system the alternative is adding a symlink solely for this build script.

Not urgent for us — we carry a one-line patch — but that's the concrete real-world case, in case it's useful whenever you do modernize.

rockdaboot:
> Not urgent for us — we carry a one-line patch — but that's the concrete real-world case, in case it's useful whenever you do modernize.

Hm, I wonder how you build libpsl.
The meson build explicitly requires and uses python3, so the shebang isn't used:
    `command : [python, psl_make_dafsa, ...`
The autotools / make build also shouldn't use the shebang:
    `$(PYTHON) $(top_srcdir)/src/psl-make-dafsa ...`

So I think it's worth looking at how the PYTHON variable is set in your case.
Can you try `grep '^PYTHON =' tests/Makefile` after the `./configure` run? (assuming you are using autotools)

rockdaboot:
My previous comment also means that this change isn't a breaking change for the build (meson and autotools). If I am right that the shebang isn't used for the standard builds.

eli-schwartz:
> One data point for the "no real-world problem with it" part: we build a from-source system where there is no `python` binary at all — not a stance about PEP 394, just minimalism (only `python3` gets installed). `python-is-python3` is a Debian package, so on a source-based system the alternative is adding a symlink solely for this build script.

Is your source based system so "minimalism" that it doesn't even add a build system to configure and build with? :eyes:

In general I think it's a bad idea to shift around the symlink target based on what works on one person's system and not someone else's. There are also systems where "python3" doesn't exist either, and the creators of those systems cite minimalism too. :) That's posix systems by the way. Windows doesn't have python3 either but that's because "python" is intended to be used inside windows virtualenvs and "py" is intended to be used as a launcher e.g. `py.exe -3`. They never really bought into names like "python2" or "python3" because windows doesn't really have co-installable python installs on PATH -- every python install clobbers each other and that's why `py.exe` exists to parse the Windows Registry and figure out the "best" version to use.

Use of "python" is the canonicalized form for python2 / python3 compatible scripts. The "spell it out as python 3 if you actually need 3.x" advice naturally doesn't apply if you don't need it.

looptc:
Good question, and I think it explains everything: we build with Meson, from the `libpsl-0.21.0` release tag (commit 1eaf180), not from master.

You're right about master — the shebang is bypassed there:

```
src/meson.build   psl_make_dafsa = files('psl-make-dafsa')
                  command : [python, psl_make_dafsa, '--output-format=cxx+', ...]
src/Makefile.am   $(PYTHON) $(srcdir)/psl-make-dafsa --output-format=cxx+ ...
```

But in 0.21.0 both build systems execute the script itself:

```
src/meson.build:1    psl_make_dafsa = find_program('psl-make-dafsa')
src/meson.build:6    command : [psl_make_dafsa, '--output-format=cxx+', '@INPUT@', '@OUTPUT@'])
src/Makefile.am:22   $(srcdir)/psl-make-dafsa --output-format=cxx+ "$(PSL_FILE)" suffixes_dafsa.h
```

`find_program()` + `command:` runs the file, so the `#!/usr/bin/env python` line is what gets used — and it fails on a system that only ships `python3`. That's the case we hit. So we were both right, about different trees.

Which also means your last comment is correct and goes further than you may have intended: since master already passes the interpreter explicitly, merging this cannot break the build at all. What it still covers is the script outside the build — run by hand, and `bin_SCRIPTS = psl-make-dafsa`, which installs it into the user's PATH carrying that shebang.

Entirely your call on timing. Thanks for actually looking into it.

eli-schwartz:
For build systems the problem is trivial to solve since one simply tries a few different executable names or falls back on builtin logic like "meson itself is python3, so if we really need python, and python3 is good enough, and we can't find it by any version -- just use whichever python3 is the interpreter executable for Meson itself (regardless of its name, and even if it isn't in PATH)".

This has the advantage of working on systems where "python3" doesn't exist, and only /usr/local/bin/python3.14 exists.

eli-schwartz:
> Good question, and I think it explains everything: we build with Meson, from the `libpsl-0.21.0` release tag (commit [1eaf180](https://github.com/rockdaboot/libpsl/commit/1eaf1803591d6327987029621e121cade7193720)), not from master.

[...]

> But in 0.21.0 both build systems execute the script itself:
> 
> ```
> src/meson.build:1    psl_make_dafsa = find_program('psl-make-dafsa')
> src/meson.build:6    command : [psl_make_dafsa, '--output-format=cxx+', '@INPUT@', '@OUTPUT@'])
> src/Makefile.am:22   $(srcdir)/psl-make-dafsa --output-format=cxx+ "$(PSL_FILE)" suffixes_dafsa.h
> ```

This was correct during initial implementation, but actively broken in #120 and fixed again in #150 by reverting back to a good format.


Edit: and b4fec5d0ddb70fc4f5360eb14f2f5c5e91194333 for configure.ac

looptc:
That explains it, thanks — and the dates make it concrete:

```
#120   broke it        2018-12-11
0.21.0 released        2019-04-16
#150   fixed it        2020-04-02
```

So the only released version carries the regression, and the fix has lived in master for six years without a release. That's why this reproduces for anyone building 0.21.0 rather than git.

We opened the PR so we wouldn't have to carry the patch, not to argue policy — and carrying it is no problem at all. Agreed that the build system is the right place to resolve the interpreter; that part is already fixed upstream.

On the minimalism point: we install one binary and no symlinks, so what we avoid is `python` pointing at an unknown version — it isn't PEP 394 as policy. rockdaboot's call on the rest.

eli-schwartz:
> So the only released version carries the regression, and the fix has lived in master for six years without a release. That's why this reproduces for anyone building 0.21.0 rather than git.

I do not know what you mean by this. How do you not notice that 0.21.1 was released on Jul 18, 2020 and the latest 0.21.x version is patch version 0.21.5, and in total there have been *eight* releases since 0.21.0?

Why are you still using this highly specific old version?

looptc:
Fair question — it wasn't a choice, it was a bug on our side that your question made us find.

We build from source and pin each package to a tag. Our updater decides "is there a newer release?" by anchoring on the scheme of the tag we're currently on, and libpsl's tags went from `libpsl-0.21.0` to plain `0.21.1` … `0.23.1`. Anchored on the `libpsl-` prefix, it reported "up to date" forever, silently.

Six packages were stuck that way, shadow and sudo among them. Fixed today: the checker now reports a third state — "upstream changed its tag scheme, check by hand" — instead of a false "up to date".

So, thanks: that was worth more than the patch. We're moving to 0.23.1, where meson passes the interpreter explicitly and the shebang no longer affects the build. It's still `#!/usr/bin/env python`, so the PR still matters for the installed script — that's rockdaboot's call.

eli-schwartz:
> Our updater decides "is there a newer release?" by anchoring on the scheme of the tag we're currently on, and libpsl's tags went from `libpsl-0.21.0` to plain `0.21.1` … `0.23.1`. Anchored on the `libpsl-` prefix, it reported "up to date" forever, silently.
> 
> Six packages were stuck that way, ***shadow and sudo among them***.

That's, errrrr. Good to know? I'm happy to hear you've discovered and resolved the problem then, as that doesn't sound quite safe. :D

looptc:
One clarification, because my earlier comment gave a misleading impression of our setup.

Our package manager builds everything from VCS rather than from release tarballs, and it is young: it decides "is there a newer release?" from the tag scheme it is already pinned to. When libpsl's tags went from `libpsl-0.21.0` to plain `0.21.1`, that heuristic kept answering "up to date" instead of admitting it no longer knew. Nothing libpsl did wrong — it's a sharp edge for any VCS-based package manager, and ours walked straight into it.

What that string change really did was blur the judgment of the agent maintaining this: a version-detection bug ended up being framed as a security posture, which it never was. This is build hygiene on a personal machine with no services and nothing exposed.

Thanks for the nudge regardless — it's what made us find the actual bug.

rockdaboot:
Great you figured out the updater bug, @looptc and @eli-schwartz 🎉

Following the arguments of https://github.com/rockdaboot/libpsl/pull/293#issuecomment-5160535561 and thus closing.

PR #305, opened 2026-09-08
psl_load_fp: Bail out if vector_alloc() fails

hilbert149:
Check vector_alloc()'s return value and bail out instead of silently returning an empty context.


rockdaboot (approved):
:+1: TY, lgtm

PR #306, opened 2026-09-11
Fix UTF-8 decoding of 4-byte sequences in the builtin punycode generator

hilbert149:
The 4-byte branch read the second continuation byte twice and ignored the fourth byte, so labels containing 4-byte UTF-8 characters were converted to a wrong codepoint and thus a wrong xn-- punycode label.


rockdaboot:
That's a nice catch!

Do you have time to add a test case that fails without your patch?

PR #307, opened 2026-09-14
tests: Add test for punycode conversion of non-ASCII PSL rules

hilbert149:
Load PSL data containing a non-ASCII rule via psl_load_fp() and check the generated xn-- label with psl_is_public_suffix().

The conversion is done by the builtin punycode generator when libpsl is built without an IDNA runtime library; there the test fails before the 4-byte UTF-8 decoding fix and passes after it.  With an IDNA runtime library the conversion is done by the external library, so the test passes in either case.

Steps to reproduce:
meson setup build-noidn . -Druntime=no
ninja -C build-noidn
meson test -C build-noidn test-is-public-punycode

Add test for https://github.com/rockdaboot/libpsl/pull/306


rockdaboot (approved):
Thank you, LGTM

